PR #6, opened 2020-06-13
Added Seqhash functions

Koeng101:
This adds seqhash functionality to the Poly library. Basically, the Seqhash function hashes any DNA/RNA/Protein sequence with the blake3 algorithm to give it a universal identifier. If the sequence is circular, like a plasmid, it first takes the lexicographically minimal string rotation using [Booth's algorithm](https://en.wikipedia.org/wiki/Lexicographically_minimal_string_rotation), then hashes the resulting sequence with the blake3 hash.


TimothyStiles:
@Koeng101 this looks awesome! I want to refactor and expand on this before merging but to do that I need two things.

1.) Can you run `go mod tiny && git add go.sum go.mod` and push those files to the prime branch of your fork? It helps with dependency tracking stuff.

2.) Can you also give me push access to your fork so I can make those changes there then merge them into the prime branch of poly? This is the best way I know of to make sure github credits you for this pull request while still allowing me to edit it before being merged.

🙏 🙏 🙏

Koeng101:
1.) Run + added + committed the command above to the prime branch of the fork. 

2.) Gave access to fork.

Koeng101:
Read comments, looks good. The reason for uppercasing is that the case never matters sequence-wise, but it impacts hashing. Could be lowercase for the same reason, but uppercase is nicer to read for sequences.

Also, the check failed because you didn't fix the `hash_test.go` file to use `seqhash` instead of `Seqhash`.

TimothyStiles:
So I just made a big push and a couple of smaller comment related commits. Everything is looking good so far but testing is still needed before this is merged.

Key things to test:

* BoothLeastRotation needs to be tested such that it's feed every rotated permutation of a circular sequence then all outputs need to be equal to eachother.

* Command usage. User experience. Make sure everything behaves as expected.

I don't see the need to test for individual hash functions. Seems tedious.

TimothyStiles:
All basic commands have been implemented and tests passed. Please review before I merge @Koeng101.

PR #11, opened 2020-06-18
Add optimization and translation functions for sequences

Koeng101:
This pull request adds codon optimization and translation functions for sequence, plus adds default values for all NCBI default codon tables. 

The general idea is that you have a `codonTable` object that stores amino acids, codons, and the number of occurrences of any given codon in proteins (defaults to 0). That `codonTable` has methods associated with it to build simple mappings between amino acids <-> codons. 

Some checks not yet added in:
- Translate requires inputs to be divisible by 3
- Creating an optimization tables requires that all codons must have at least 1 occurrence

Integration to think about:
- How do we generate `codonTable` objects from genbank files? How can we save them as JSON for use later?

@TimothyStiles please review and comment with anything you think I should add into the pull request.


transformations.go
```diff
@@ -0,0 +1,142 @@
...
+	translationTable := codonTable.generateTranslationTable()
+	length := len(nucleotides) / 3 // Assumes input sequences are divisible by 3
+	for i := 0; length > i; i++ {
+		aminoAcids += translationTable[nucleotides[i*3:(i+1)*3]]
+	}
+	return aminoAcids
+}
+
+// Function to generate default codon tables from NCBI https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi
+func generateCodonTable(aas, starts string) CodonTable {
```
TimothyStiles, line 79: What does aas stand for here? Amino acid string?

Koeng101: It stands for amino acids. On the link, they just name it "AAs", so I just copied their nomenclature (also why starts is named starts)

TimothyStiles:
@Koeng101 do you have any idea on how we would generate codon tables for arbitrary sequences? Are there any examples that we can cite/work off of?

Koeng101:
@TimothyStiles Generation of codon tables for arbitrary sequences can be done by counting the codon occurrences for each CDS feature in that GenBank file. This is one of the reasons we were working on the "location" feature previously.

transformations.go
```diff
@@ -0,0 +1,238 @@
...
+	return codonFrequencyHashMap
+}
+
+// Generate map of codons -> amino acid
+func (codonTable CodonTable) generateTranslationTable() map[string]string {
+	var translationMap = make(map[string]string)
+	for _, aminoAcid := range codonTable.AminoAcids {
+		for _, codon := range aminoAcid.Codons {
+			translationMap[codon.Triplet] = aminoAcid.Letter
+		}
```
tylermaran, line 139: Maybe I don't know go very well. But is this a `for loop` in a `for loop` in a `map` >.>

transformations.go
```diff
@@ -0,0 +1,238 @@
...
+			sequenceBuffer.WriteString(feature.getSequence())
+		}
+	}
+	optimizationTable := codonTable.generateOptimizationTable(sequenceBuffer.String())
+	for _, aminoAcid := range aminoAcids {
+		codons.WriteString(optimizationTable[string(aminoAcid)].Pick().(string))
+	}
+	return codons.String()
+}
+
```
Koeng101, line 68: Optimization table should be an input to the `Optimize` function. This is because the AnnotatedSequences may be huge (in case of human chromosomes), and you will need to load multiple files in order to get the correct optimization table (again, with multiple human chromosomes). Once generated, you won't need to do that again.

Ideally, there would be some kind of robust import / export function to use JSON codon tables. But at minimum, split this function so that can be added later.

Koeng101 (commented):
One function should be split up because AnnotatedSequences can get big!

Koeng101 (commented):
Looks good to me. I'll make a git issue of JSON representations for the codon table stuff. <hold on, found a bug>

TimothyStiles:
Alright this PR is getting to be a monster. I refactored the way command line commands are tested for easier debugging and now have two simple command line utilities that both translate and optimize streams of sequence strings. It also has related library functions as well.

I'm tired and I'm merging it y'all.

🎉 🎉 🎉

Issue #24, opened 2020-07-03
No sequence parsing for genbank files exported from Snapgene


Koeng101:
**Describe the bug**
Genbank files exported by Snapgene have no sequence output in their JSON files.

**To Reproduce**
[puc19.gbk](http://ix.io/2qNF)
[puc19_snapgene.gb](http://ix.io/2qNO)
[puc19.json](http://ix.io/2qNR)
[puc19_snapgene.json](http://ix.io/2qNP)

```bash
$ cat puc19.gbk | poly c -i gb -o json > puc19.json
$ cat puc19_snapgene.gb | poly c -i gb -o json > puc19_snapgene.json
cat puc19.json | jq ".Sequence.Sequence"
$ cat puc19.json | jq ".Sequence.Sequence"
[...]

PR #26, opened 2020-07-03
Added Sequence method to Feature object

Koeng101:
This pull request adds a minimally viable way to get the sequence of features. This can be used to fill codon tables for optimization, or to just get the DNA sequence of features. 

Testing is currently blocked with bug #24 .

Definitely needs some improvement, but as-is it works for the following from this page (basically, all the Location types that are relevant for CDSs): http://www.insdc.org/files/feature_table.html#3.4.3

```
340..565
join(12..78,134..202)
complement(34..126)
complement(join(2691..4571,4918..5163))
join(complement(4918..5163),complement(2691..4571))
 ```

Usage:
```
f := feature.Sequence(annotatedSequence)
```


TimothyStiles:
@Koeng101 please review then I'll review again and merge!

PR #27, opened 2020-07-06
fixed origin parsing bug from issue #24.

TimothyStiles:
This patch fixes an origin parsing bug related to whitespace when genbank files are output by snapgene as discussed in issue #24. It includes a regression test, and a test file.


.gitignore
```diff
@@ -1,2 +1,3 @@
 data/genbank
 public
+node_modules
```
Koeng101, line 3: Why is this added?

Koeng101: un-tracked files in docs directory fix.

PR #31, opened 2020-07-11
Add JSON Schema to Poly

Koeng101:
A JSON Schema is a cross-language validation format. If we had a JSON Schema for Poly's Genbank output, in the future it would be much easier to get adoption of our JSON-gene format, since anyone can test their Genbank parser + JSON output + Genbank writer against this schema format. In addition, it would give guarantees to downstream applications consuming the JSON, which is nice for application developers. 

This pull request contains my first pass at a JSON Schema, and a test function in `io_test.go` to validate that the schema validates output JSON. The JSON Schema will need improvements over time, but this gives a decent first pass and something to iterate on.

### DIFF
--- a/data/polySchema.json
+++ b/data/polySchema.json
@@ -25,7 +25,7 @@
 			"Locus": {"type": "object", "properties": {
 				"Name": {"type": "string"},
 				"SequenceLength": {"type": "string"},
-				"MoleculeType": {"type": "string", "enum": ["DNA","RNA"]},
+				"MoleculeType": {"type": "string", "enum": ["DNA", "genomic DNA", "genomic RNA", "mRNA", "tRNA", "rRNA", "other RNA", "other DNA", "transcribed RNA", "viral cRNA", "unassigned DNA","unassigned RNA"]},
 				"GenBankDivision": {"type": "string"},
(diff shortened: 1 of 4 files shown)


data/polySchema.json
```diff
@@ -0,0 +1,72 @@
...
+			"Accession": {"type": "string"},
+			"Version": {"type": "string"},
+			"Keywords": {"type": "string"},
+			"Organism": {"type": "string"},
+			"Source": {"type": "string"},
+			"Origin": {"type": "string"},
+			"Locus": {"type": "object", "properties": {
+				"Name": {"type": "string"},
+				"SequenceLength": {"type": "string"},
+				"MoleculeType": {"type": "string", "enum": ["DNA","RNA"]},
```
TimothyStiles, line 28: There's a ton of [official molecule types](https://github.com/TimothyStiles/poly/blob/0b37805da7c94831109b62298868be891859e411/io.go#L360) in io.go.

Koeng101: Got it, and added.

Koeng101:
This'll be changing until Poly version 1.0, so closing this for now.

Issue #10, opened 2020-06-17
Fasta file format IO


TimothyStiles:
**Describe the solution you'd like**
I'd like a fasta parser than parses fasta into an AnnotatedSequence struct or set of AnnotatedSequence structs. Most of the struct can be blank but the AnnotatedSequence.Sequence.Sequence should be a single string of characters with no whitespace and AnnotatedSequence.Sequence.Description should be the description. This is a really great first contribution and there's no rush to implement so I thought it'd be great to share here!

**Additional context**
Here's a [spec](https://en.wikipedia.org/wiki/FASTA_format) from wikipedia. The resulting parser should be placed in [io.go](https://github.com/TimothyStiles/poly/blob/prime/io.go) along with tests in [io_test.go](https://github.com/TimothyStiles/poly/blob/prime/io_test.go). More information on the AnnotatedSequence struct can also be found in [io.go](https://github.com/TimothyStiles/poly/blob/prime/io.go).

0x106:
@TimothyStiles I'd be happy to work on this feature if you'd like?

TimothyStiles:
@0x106 go for it! Is there anything I can help with?

0x106:
I should be fine to get started thanks. I had a look through io.go as a starting point and am going to set up a bunch of test cases before I actually write anything.

0x106:
Hey I pushed an initial version. It reads sequences from a .fasta file into an array of AnnotatedSequence structs (header line gets written into `Sequence.Description` and sequence into `Sequence.Sequence`.

This was my first time writing anything in Go so you might want to give it a quick check to make sure it's idiomatic / actually correct.

TimothyStiles:
Thanks to @0x106 for adding this in PR #37! There still needs to be a build fasta function but we can merge before adding that.

PR #37, opened 2020-08-24
Read a FASTA file containing multiple sequences, extract the definition and sequence from each.

0x106:
Fasta files contain multiple sequences, delimited by an empty line. Each sequence contains a header line that begins with '>' and then multiple lines of sequence data.

This pull request assumes that there are one or more sequences per file, and reads them into an array. `Sequence.Definition` and `Sequence.Sequence` are populated per sequence. 

The sequences themselves are not validated, and nor are the header lines.


TimothyStiles:
@0x106 this looks awesome. The only thing I'd do is create a separate BuildFasta function from WriteFasta but these changes can be squashed and merged before then. I'll also close out issue #10.

PR #45, opened 2020-10-22
add windows to test matrix

TimothyStiles:
A side effect of #11 being merged was that I refactored the command tests and they should now be able to run on windows. This pull request tests that theory.


TimothyStiles:
So I was super wrong on that. It looks like windows just don't like unix style paths and I'm not quite sure what to do.

TimothyStiles:
Turns out it isn't GFFIO that is broken but literally all IO for windows panics. More research needs to be done.

TimothyStiles:
Turns out windows users can use some official Microsoft subsystem to avoid powershell so we don't have to support windows explicitly :D

PR #61, opened 2020-10-31
Refactored parsing functions to accept bytes instead of string

TimothyStiles:
I've refactored parsing functions to accept bytes instead of strings. Most use cases ended up converting bytes to string before input.


TimothyStiles:
This was solved long ago in a different PR and I forgot to close this one.

PR #65, opened 2020-11-01
Coverage

TimothyStiles:
On the quest to get 100% code coverage. Library is almost there. The command line utility tho....


TimothyStiles:
At this stage most of the library besides a few continues, breaks, and conditionals in io.go is covered. The Hash command needs to be refactored to have simpler outputs which should help with coverage there. Merging and will follow up with new refactor-commands branch.

PR #71, opened 2020-11-17
Fix error weightedrand.NewChooser returns 2 values

Koeng101:
On a fresh install, I got the following error - 

`../../../pkg/mod/github.com/!timothy!stiles/poly@v0.6.1/transformations.go:200:34: assignment mismatch: 1 variable but weightedrand.NewChooser returns 2 values`

It appears 19 days ago, weightedrand updated to handle error conditions (https://github.com/mroth/weightedrand/commit/953da99f62a7995b5b73a21940be7af2a0be59cc), which broke the code. This merge updates `go.mod`, `go.sum`, and `transforms.go` to fix this error.


Koeng101:
This is stupid, so I'm just going to not update that library.

PR #79, opened 2020-11-23
Add support for Gz'd flat files

Koeng101:
the FTP genbank flat files are distributed in files compressed with GZ. This allows you to directly read those files with `ReadGbkFlatGz`.


io.go
```diff
@@ -1469,6 +1470,19 @@ func ReadGbkFlat(path string) []Sequence {
 	return sequences
 }
 
+// ReadGbkFlatGz reads flat gzip'd genbank files, like the ones provided by the NCBI FTP server
+func ReadGbkFlatGz(path string) []Sequence {
+	file, err := ioutil.ReadFile(path)
```
TimothyStiles, line 1475: If you're not using err for anything just replace it with an underscore.

io.go
```diff
@@ -1459,7 +1460,7 @@ func ReadGbkMulti(path string) []Sequence {
 	return sequences
 }
 
-// ReadGbkFlat reads flat genbank files, like the ones provided by the NCBI FTP server
+// ReadGbkFlat reads flat genbank files, like the ones provided by the NCBI FTP server (after decompression)
 func ReadGbkFlat(path string) []Sequence {
 	file, err := ioutil.ReadFile(path)
 	if err != nil {
```
TimothyStiles, line 1466: Is this only returning if err != nil? return statements aren't supposed to be in control flow stuff. Normally you'd just use control flow to determine what to set the return variable as and then return it at the end of the function. This can mess with some linters otherwise.

TimothyStiles (changes requested):

io.go
Koeng101: Fixed.

io.go
Koeng101: Fixed.

PR #83, opened 2020-12-06
Implementing concurrent FASTA parsing

Koeng101:
This is specifically in order to parse Uniprot's FASTA databases - they do not fit in RAM, so you need a way to concurrently parse the file while you are reading it. This PR implements that feature.

## What it do

It can currently parse Uniprot Sprot in 3.1s, or at a speed of about 27.4mb/s (of a compressed GZ file). With that speed, it could parse Uniprot TrEMBL in 1678s, or about ~28 minutes (but maybe faster, idk, haven't tried it on TrEMBL). 

Though, tbh, this isn't a lot faster than Biopython. I tested to see how fast Biopython was with the code below:

```python
import gzip
from Bio import SeqIO
import time

tic = time.perf_counter()

with gzip.open("uniprot_sprot.fasta.gz", "rt") as handle:
    for record in SeqIO.parse(handle, "fasta"):
        name = record.id
toc = time.perf_counter()
print(f"Parsed Sprot in {toc - tic:0.4f} seconds")
```
"Parsed Sprot in 4.6871 seconds"

Though the [Biopython code](https://github.com/biopython/biopython/blob/master/Bio/SeqIO/FastaIO.py) isn't concurrent, which may have consequences down the line when you actually want to do stuff with any data you're getting.

## What I changed

I did not use the current FASTA reader at all, opting instead to reimplement it so that it works better on file streams. We should probably think about how to reconcile all that. In addition, I did NOT use the Sequence struct - I used a custom FASTA structure that was much simpler. How do we reconcile that? Do we want to only use Sequence structs, should we use a minimized version, or a function that can switch between the two?

## What needs to be changed before pull

@TimothyStiles I'd like to know what you think about the code and how / where I should format it into (right now it just exists between the JSON and FASTA code blocks in `io.go` and `io_test.go` (didn't add it into FASTA block cause there is other code there)

Cheers,

Keoni


io_test.go
```diff
@@ -430,6 +442,7 @@ func ExampleReadGbkFlat() {
 }
 
 func ExampleReadGbkFlatGz() {
+	// sequences := ReadGbkFlatGz("data/gbbct358.seq.gz")
```
TimothyStiles, line 445: Remove this?

TimothyStiles:
@Koeng101 how fast does this go compared to biopython exactly?

Koeng101:
@TimothyStiles For the parsing of Uniprot Sprot:

Poly: 3.1s
Biopython: 4.6s

Koeng101:
@TimothyStiles Your bot is bugged again. Just check out the codecov comment - inconsistent data.

Btw, there should be a full test for the smallfasta file that all parts are properly parsed and parsed once (had some errors where the start line would be two different objects). Would love to get some help writing that, since it is pretty simple.

We should also discuss what we want to do with the other fasta parsing functionality - it works, and it is fantastic we got a pull, but how do we separate that work and this work?

Koeng101:
@TimothyStiles I code climate ain't telling me the 6 issues I need to fix, when I click on it it gives me 106 fixes that are from the prime branch.

Koeng101:
Closing due to #113 being superior.

PR #85, opened 2020-12-17
Added CompromiseCodonTable to Poly

Koeng101:
Adds a single function called `CompromiseCodonTable` to transformations.go

This function takes 2 codon tables and compromises them together. Very useful for creating compromise tables between E.coli and B.subtilis for example (I would like to use this for some genes I am optimizing for formate bioutilization).

### DIFF
--- a/transformations.go
+++ b/transformations.go
@@ -469,6 +469,7 @@ Keoni
 // CompromiseCodonTable takes 2 CodonTables and makes a new CodonTable
 // that is an equal compromise between the two tables.
 func CompromiseCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable, cutOff float64) (CodonTable, error) {
+	// Initialize output CodonTable, c
 	var c CodonTable
 	// Check if cutOff is too high or low (this is converted to a percent)
 	if cutOff < 0 {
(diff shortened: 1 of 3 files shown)


transformations.go
```diff
@@ -435,3 +436,130 @@ func WriteCodonJSON(codontable CodonTable, path string) {
...
+since these are all already balanced).
+
+Godspeed,
+
+Keoni
+******************************************************************************/
+
+// CompromiseCodonTable takes 2 CodonTables and makes a new CodonTable
+// that is an equal compromise between the two tables.
+func CompromiseCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable, cutOff float64) (CodonTable, error) {
```
TimothyStiles, line 471: I understand the purpose of this function but I don't really understand how it works. Could you clean up some names and add comments?

transformations.go
```diff
@@ -435,3 +436,130 @@ func WriteCodonJSON(codontable CodonTable, path string) {
...
+func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) CodonTable {
+	var c CodonTable
+
+	// Take start and stop strings from first table
+	c.StartCodons = firstCodonTable.StartCodons
+	c.StopCodons = firstCodonTable.StopCodons
+
+	// Add up codons
+	var finalAminoAcids []AminoAcid
+	for _, firstAa := range firstCodonTable.AminoAcids {
```
TimothyStiles, line 550: There's some nested pyramid of death stuff going on here. What's going on? Can we simplify this?

TimothyStiles (dismissed):

transformations.go
Koeng101: It is adding all the weights of triplets together. With the current implementation of the CodonTables, there isn't really a way to simplify this.

transformations.go
Koeng101: Added more comments. I think the names are pretty good in this one, but please tell me if any are unclear. I think the comments might clear things up. (changed a few things so it would be easier to reason about in latest commit, tests still pass and everything and the code acts the same)

TimothyStiles:
Review and looks good. Some variables could use better names but overall it's readable. Merging.

PR #86, opened 2020-12-19
Fixed panics during large file parsing

Koeng101:
This PR fixes the parsing issues when it comes to big' ole real-life GenbankFlatFiles.

Note: I have no idea WHY exactly it fixed them all. Basically, all I did was add length checks in 4 places:

`quickMetaCheck` (if len(line) == 0, flag is false)
`quickSubMetaCheck` (if len(line) == 0, flag is false)

check if len(startEndSplit) is > 1 before doing `end, _ = strconv.Atoi(partialRegex.ReplaceAllString(startEndSplit[1], ""))`
finally, check if there is still line to parse over when trying to find paratheses in `switch expression[firstInnerParentheses+i]`. Apparently, sometimes there ain't?

And commented out a line of code
`attributeValue = strings.TrimSpace(attributeSplit[1])`
I am not sure why this line is added - it is completely untested, and if you actually encounter it the system panics. 

In particular, I tested with the following genbank file https://siasky.net/AABxAejn_kkw0H5O8ECVdPhFuMXy1uib_Z8B5JE5DctSuQ

It is pretty large, so I didn't actually push it to the main branch. It also takes quite a few seconds to parse, so there is that too.


Koeng101:
Forgot the bot was an a-hole, let me try to get those tests....

Koeng101:
@TimothyStiles I think your bot is bugged. `93.54% <0.00%> (-6.46%)` doesn't make very much sense.

PR #87, opened 2020-12-19
UniqueSequence function

Koeng101:
This pull adds a function called "UniqueSequence", which is an interesting little function that uses a unique algorithm to accomplish its goal. It is useful for creating barcodes on primers and is useful for creating CDSs with maximally distant coding sequences.

If you call the UniqueSequence function, you want sequences of a certain length with a maximum subsequence length, so that all output sequences do not have any shared subsequence of the maximum subsequence length. In addition, you may pass in strings or functions on strings to ban certain sequences (such as a restriction enzyme site or high GC content). This removes any sequence that doesn't pass those checks, while still outputting as many as possible which pass. 

To accomplish this, we create a [de Bruijn Sequence](https://en.wikipedia.org/wiki/De_Bruijn_sequence). To quote - "a de Bruijn sequence of order n on a size-k alphabet A is a cyclic sequence in which every possible length-n string on A occurs exactly once as a substring".

This exact matching is important - it means that any length of n on a debruijn sequence will be unique. Because of that property, we can iterate over debruijn sequences, and every iteration will be a unique sequence, without any need to check the other sequences. 

This function is specifically useful for creating barcodes, since it ensures that you will have uniquely generated sequences, while being non-random and enabling easy sorting of those barcodes. This function could also be applied to making unique CDS sequences.

I need this function to generate unique barcodes for Nanopore sequencing of materials.


Koeng101:
A note on efficiency and purity - the "best" way to implement this would be approximately a traveling salesman problem in which each subsequence below the minimal subsequence is checked between each possible combination of Xmer. 

IMO that is kind of overkill, and this function should be more than enough for barcoding purposes.

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
...
 }
+
+/******************************************************************************
+Start of the De Bruijn stuff
+******************************************************************************/
+
+// DeBruijn generates a DNA DeBruijn sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func DeBruijn(n int) string {
```
TimothyStiles, line 115: @Koeng101 should this be a public function?

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
 	meltingTemp, _, _ := SantaLucia(sequence, primerConcentration, saltConcentration, magnesiumConcentration)
 	return meltingTemp
 }
+
+/******************************************************************************
+Start of the De Bruijn stuff
+******************************************************************************/
+
+// DeBruijn generates a DNA DeBruijn sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
```
TimothyStiles, line 113: Can you add this reference link and write a thing in line on De Bruijn sequences an how they work? https://en.wikipedia.org/wiki/De_Bruijn_sequence

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
...
+// Pulled and adapted from here
+func DeBruijn(n int) string {
+	k := 4
+	alphabet := "ATGC"
+	a := make([]byte, k*n)
+	var seq []byte
+	var db func(int, int) // recursive closure
+	db = func(t, p int) {
```
TimothyStiles, line 121: Can you change all single letter variables to whole words that are readable?

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
...
+	for _, i := range seq {
+		buf.WriteByte(alphabet[i])
+	}
+	b := buf.String()
+	return b + b[0:n-1] // as cyclic append first (n-1) digits
+}
+
+// UniqueSequence takes a channel and fills it with Unique sequences
+// which can then be sorted for desired properties (high/low GC,
+// restriction enzyme sites, etc)
```
TimothyStiles, line 146: Can you make an inline explanation to what this does on a function level?

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
...
+func UniqueSequence(c chan string, length int, maxSubSequence int, banned []string, bannedFunc []func(string) bool) {
+	var start int
+	var end int
+	debruijn := DeBruijn(maxSubSequence)
+	for i := 0; (i*(length-(maxSubSequence-1)))+length < len(debruijn); {
+		start = i * (length - (maxSubSequence - 1))
+		end = start + length
+		i++
+		for _, b := range banned {
```
TimothyStiles, line 155: ranging through banned sequences.

primers.go
```diff
@@ -103,3 +104,79 @@ func MeltingTemp(sequence string) float64 {
...
+					close(c)
+					return
+				} else {
+					start++
+					end++
+					i++
+				}
+			}
+		}
+		c <- debruijn[start:end]
```
TimothyStiles, line 179: what does this do?

primers_test.go
```diff
@@ -80,3 +81,50 @@ func TestMeltingTemp(t *testing.T) {
...
+	var output string
+	for a := range c {
+		output = a
+	}
+
+	fmt.Println(output)
+	// Output: TCTCGGTCGCTCCGTCCCGG
+}
+
+func TestUniqueSequence(t *testing.T) {
```
TimothyStiles, line 104: Why are channels involved and why can't this single threaded?

TimothyStiles (changes requested):

primers.go
NickNolan, line 146: If I'm understanding this correctly:
 - `DeBruijn()` just generates a string such that every possible subsequence of length `n` is both represented and unique in the string;
 - `UniqueSequence()` generates the set `c` of strings of length `length` such that each subsequence of length `maxSubSequence` is unique within the original string. It removes any items in the set that are also contained in `banned` or that satisfy `bannedFunc()`.

Is my understanding here correct at a functional level?

primers_test.go
```diff
@@ -80,3 +81,50 @@ func TestMeltingTemp(t *testing.T) {
...
+func ExampleUniqueSequence() {
+	c := make(chan string)
+	go UniqueSequence(c, 20, 4, []string{"CTCTCGGTCGCTCC"}, []func(string) bool{})
+	var output string
+	for a := range c {
+		output = a
+	}
+
+	fmt.Println(output)
+	// Output: TCTCGGTCGCTCCGTCCCGG
```
dexho, line 101: Is there a reason why there is just one sequence outputted? What about "sorting for high/low GC, restriction enzyme sites" after using the channel to generate multiple sequences?

primers.go
Koeng101: Gotcha. Added.

primers.go
Koeng101: Yes.

primers.go
Koeng101: I would, but I don't really know the proper names for all of them or precisely what they do

primers.go
Koeng101: I think I added that at top

primers.go
Koeng101: yes

primers.go
Koeng101: Added in comments

Koeng101 (commented):
Resolved problems + some new code that should make things clearer

primers.go
TimothyStiles: Could you write a comment since it's exported so that it shows up in our docs?

primers.go
TimothyStiles: Can you or someone find out?

primers_test.go
TimothyStiles: @dexho can you respond to this and @Koeng101 can you maybe put a comment in this file about it?

TimothyStiles:
> A note on efficiency and purity - the "best" way to implement this would be approximately a traveling salesman problem in which each subsequence below the minimal subsequence is checked between each possible combination of Xmer.
> 
> IMO that is kind of overkill, and this function should be more than enough for barcoding purposes.

@Koeng101 how annoying would it be to upgrade to the "best" way in the future.

Koeng101:
Ya know, I might be wrong about that being most efficient. Here is the logic:

- In the ideal case, you could take take the intersection of two strings of length X. The length of this intersection would be the max length shared between those two strings. A naive n^2 algorithm would compare every intersection to every intersection to find optimal pairings
- However, by the definition of the de Bruijn sequences, every subsequence of a certain length Y will be contained within DeBruijn(Y). Therefore, the `len(DeBruijn(Y))` will be the maximum length possible to find only sequences with subsequence shared of length Y-1. 
- So, if you want to minimize the length of identical sequence between strings of length X in DeBruijn(Y), you would actually just iterate on DeBruijn(Y-n) until you can't find enough sequences. Your optimal DeBruijn will occur right before that, I guess at DeBruijn(Y-n+1).

Of course, that changes when there are other functions you need to constrain on (for example, when you're not actually looking for unique sequences, but for sequences under certain constraints). The change therefore would be to add in something that automatically iterates down `DeBruijn(Y-n)`, which shouldn't be too hard.

Koeng101:
I fixed all the things specific to this pull request by code climate except the "complexity" issues. A review would be appreciated @TimothyStiles

primers.go
```diff
@@ -103,3 +104,80 @@ func MeltingTemp(sequence string) float64 {
...
+/******************************************************************************
+Start of the De Bruijn stuff
+******************************************************************************/
+
+// DeBruijn generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func DeBruijn(n int) string {
+	k := 4
```
TimothyStiles, line 117: what is k?

primers.go
TimothyStiles: @Koeng101 still got single letter variables. Have no idea what's going on.

primers.go
```diff
@@ -103,3 +104,80 @@ func MeltingTemp(sequence string) float64 {
 	meltingTemp, _, _ := SantaLucia(sequence, primerConcentration, saltConcentration, magnesiumConcentration)
 	return meltingTemp
 }
+
+/******************************************************************************
+Start of the De Bruijn stuff
+******************************************************************************/
+
```
TimothyStiles, line 111: Give biocontext and why this is useful enough to be public.

primers_test.go
```diff
@@ -80,3 +81,51 @@ func TestMeltingTemp(t *testing.T) {
...
+func ExampleDeBruijn() {
+	a := DeBruijn(4)
+
+	fmt.Println(a)
+	// Output: AAAATAAAGAAACAATTAATGAATCAAGTAAGGAAGCAACTAACGAACCATATAGATACATTTATTGATTCATGTATGGATGCATCTATCGATCCAGAGACAGTTAGTGAGTCAGGTAGGGAGGCAGCTAGCGAGCCACACTTACTGACTCACGTACGGACGCACCTACCGACCCTTTTGTTTCTTGGTTGCTTCGTTCCTGTGTCTGGGTGGCTGCGTGCCTCTCGGTCGCTCCGTCCCGGGGCGGCCGCGCCCCAAA
+}
+
+func ExampleUniqueSequence() {
+	uniqueSequences := make(chan string)
+	go UniqueSequence(uniqueSequences, 20, 4, []string{"CTCTCGGTCGCTCC"}, []func(string) bool{})
```
TimothyStiles, line 94: Any reason you're using go routines and channels in this test?

TimothyStiles (changes requested):

primers.go
jkhales: single letter variables are also really tough to grep and refactor.

primers.go
jkhales: And why is it 4? Is this the length of the `alphabet` variable below?

primers.go
```diff
@@ -103,3 +104,80 @@ func MeltingTemp(sequence string) float64 {
...
+// DeBruijn generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
```
jkhales, line 115: Is there anyway we can just import a function directly from this golang library you mentioned?

https://rosettacode.org/wiki/De_Bruijn_sequences#Go

writing our own means we maintaining and testing our own.

primers.go
```diff
@@ -103,3 +104,80 @@ func MeltingTemp(sequence string) float64 {
...
+		buf.WriteByte(alphabet[i])
+	}
+	b := buf.String()
+	return b + b[0:n-1] // as cyclic append first (n-1) digits
+}
+
+// UniqueSequence takes a channel and fills it with Unique sequences
+// which can then be sorted for desired properties (high/low GC,
+// restriction enzyme sites, etc)
+func UniqueSequence(c chan string, length int, maxSubSequence int, banned []string, bannedFunc []func(string) bool) {
```
jkhales, line 148:
```suggestion
func UniqueSequence(channel chan string, length int, maxSubSequence int, banned []string, bannedFunc []func(string) bool) {
```

Single letter variable names should be avoided except possibly in very short anonymous functions, and even then it's debatable.

jkhales, line 148:
```suggestion
func CreateBarcode(c chan string, length int, maxSubSequence int, banned []string, bannedFunc []func(string) bool) {
```

UniqueSequence is a too generic of a name to be informative. 

If this has uses other than Barcodes, than maybe `CreateChannelOfNonOverLappingDeBruijnSequences` or something? 

I like variable names that give a lot of information and do double duty as comments.

primers_test.go
jkhales: Are channels necessary in general for this function? Is it possible this is over optimization?

primers_test.go
```diff
@@ -80,3 +81,51 @@ func TestMeltingTemp(t *testing.T) {
...
+	var output string
+	for uniqueSeq := range uniqueSequences {
+		output = uniqueSeq
+	}
+
+	// Only we only want 1 output to check against
+	fmt.Println(output)
```
jkhales, line 101:
```suggestion
	fmt.Println(<-uniqueSequences)
```
If we just check the first item, we can avoid the for loop here.

primers_test.go
```diff
@@ -80,3 +81,51 @@ func TestMeltingTemp(t *testing.T) {
...
+}
+
+func TestUniqueSequence(t *testing.T) {
+	c := make(chan string)
+	testFunc := func(s string) bool {
+		if strings.Contains(s, "GGCCGCGCCCC") {
+			return false
+		} else {
+			return true
+		}
```
jkhales, line 112:
```suggestion
		return  !strings.Contains(s, "GGCCGCGCCCC")
```

primers.go
jkhales:
```suggestion
	alphabetSize := alphabet.length
```

primers.go
```diff
@@ -103,3 +104,80 @@ func MeltingTemp(sequence string) float64 {
...
+	alphabet := "ATGC"
+	a := make([]byte, k*n)
+	var seq []byte
+	var db func(int, int) // recursive closure
+	db = func(t, p int) {
+		if t > n {
+			if n%p == 0 {
+				seq = append(seq, a[1:p+1]...)
+			}
+		} else {
```
jkhales, line 127:
```suggestion
		if t > n && n%p == 0 {
				seq = append(seq, a[1:p+1]...)
		} else {
```

primers_test.go
Koeng101: Have you tried running that exact code? I get a panic when I do

primers.go
Koeng101: I do not know why. But this block of code does not work and causes failure in run time.

primers.go
Koeng101: Fixed. I agree with naming scheme.

primers.go
Koeng101: Unfortunately, there is not. It is not a Golang library, it is just a repository of ways to do basic algorithms in different languages.

primers_test.go
Koeng101: Quite possibly over optimization. At the time of writing, I was thinking it would be a bottleneck, but after some experimentation, I found this not to be the case. Let me see if I can simplify.

Koeng101: Resolved. Removed over optimization.

primers.go
Koeng101: Yep, it was based on length of alphabet.

Koeng101: Fixed

primers.go
Koeng101: Moving conversation down.

primers.go
Koeng101: Will do with biocontext.

The single letter variables are there because I don't really know what it is doing either (I just took examples from RosettaCode). I should figure out that portion myself.

primers.go
```diff
@@ -103,3 +104,78 @@ func MeltingTemp(sequence string) float64 {
...
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func DeBruijn(n int) string {
+	alphabet := "ATGC"
+	k := len(alphabet)
+	a := make([]byte, k*n)
+	var seq []byte
+	var db func(int, int) // recursive closure
+	db = func(t, p int) {
```
jkhales, line 122: p is never mutated in this recursion. And the anoymous function is only ever called with p=1.

I would make this function take only a single variable and set `p` outside of it and with an explicit name that demystifies it.

jkhales, line 122:
```suggestion
	ConstructDeBruijn = func(t, p int) {
```

primers.go
```diff
@@ -103,3 +104,78 @@ func MeltingTemp(sequence string) float64 {
...
+
+/******************************************************************************
+Start of the De Bruijn stuff
+******************************************************************************/
+
+// DeBruijn generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func DeBruijn(n int) string {
```
jkhales, line 116:
```suggestion
func NucleobaseDeBruijnSequence(n int) string {
```

jkhales (approved):
This is looking pretty good to me. Left a few last suggestions inline.

Koeng101:
Alright, I do not know what `t` and `p` really mean. I've been searching rosetta code, looking at wikipedia, etc. Thoughts @TimothyStiles

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
 	meltingTemp, _, _ := SantaLucia(sequence, primerConcentration, saltConcentration, magnesiumConcentration)
 	return meltingTemp
 }
+
+/******************************************************************************
```
TimothyStiles, line 108: Explain more about DNA barcoding. How does it work biologically? Where else besides nanopore stuff do you see it? Non-biologists won't understand much context here.

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+}
+
+// CreateBarcodes creates a list of barcodes given a desired barcode length, the maxSubSequence shared in each barcode,
+// any banned sequences within the barcode, and any functions which may ban sequences within the barcode.
+func CreateBarcodes(length int, maxSubSequence int, bannedSequences []string, bannedFunctions []func(string) bool) []string {
+	var barcodes []string
+	var start int
+	var end int
+	debruijn := NucleobaseDeBruijnSequence(maxSubSequence)
+	for i := 0; (i*(length-(maxSubSequence-1)))+length < len(debruijn); {
```
TimothyStiles, line 173: What are you iterating for `i`? Name variable that instead.

primers_test.go
```diff
@@ -80,3 +81,34 @@ func TestMeltingTemp(t *testing.T) {
 		t.Errorf("MeltingTemp has changed on test. Got %f instead of %f", calcTM, expectedTM)
 	}
 }
+
+func ExampleNucleobaseDeBruijnSequence() {
+	a := NucleobaseDeBruijnSequence(4)
```
TimothyStiles, line 86: What is 'a'?

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+
+Keoni
+
+******************************************************************************/
+
+// NucleobaseDeBruijnSequence generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func NucleobaseDeBruijnSequence(substringLength int) string {
```
TimothyStiles, line 137: This function needs way more comments aside from `t, p, and a` being renamed.

primers_test.go
```diff
@@ -80,3 +81,34 @@ func TestMeltingTemp(t *testing.T) {
...
+
+	fmt.Println(a)
+	// Output: AAAATAAAGAAACAATTAATGAATCAAGTAAGGAAGCAACTAACGAACCATATAGATACATTTATTGATTCATGTATGGATGCATCTATCGATCCAGAGACAGTTAGTGAGTCAGGTAGGGAGGCAGCTAGCGAGCCACACTTACTGACTCACGTACGGACGCACCTACCGACCCTTTTGTTTCTTGGTTGCTTCGTTCCTGTGTCTGGGTGGCTGCGTGCCTCTCGGTCGCTCCGTCCCGGGGCGGCCGCGCCCCAAA
+}
+
+func ExampleCreateBarcodes() {
+	barcodes := CreateBarcodes(20, 4, []string{"CTCTCGGTCGCTCC"}, []func(string) bool{})
```
TimothyStiles, line 93: Passing a slice of string functions that return a bool? What's going on with this?

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+	alphabet := "ATGC"
+	alphabetLength := len(alphabet)
+	a := make([]byte, alphabetLength*substringLength)
+	var seq []byte
+	var ConstructDeBruijn func(int, int) // recursive closure
+	ConstructDeBruijn = func(t, p int) {
```
TimothyStiles, line 143: Comment this control flow as much as possible. Did this need to be defined as a function within this function's scope?

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+// NucleobaseDeBruijnSequence generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func NucleobaseDeBruijnSequence(substringLength int) string {
+	alphabet := "ATGC"
+	alphabetLength := len(alphabet)
+	a := make([]byte, alphabetLength*substringLength)
+	var seq []byte
+	var ConstructDeBruijn func(int, int) // recursive closure
```
TimothyStiles, line 142: Oh boy what up recursion. Can you explain why this is necessary in the comments?

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+******************************************************************************/
+
+// NucleobaseDeBruijnSequence generates a DNA DeBruijn sequence with alphabet ATGC. DeBruijn sequences are basically a string with all unique substrings of an alphabet represented exactly once.
+// https://en.wikipedia.org/wiki/De_Bruijn_sequence
+// https://rosettacode.org/wiki/De_Bruijn_sequences#Go
+// Pulled and adapted from here
+func NucleobaseDeBruijnSequence(substringLength int) string {
+	alphabet := "ATGC"
+	alphabetLength := len(alphabet)
+	a := make([]byte, alphabetLength*substringLength)
```
TimothyStiles, line 140: What is 'a' here?

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+	var ConstructDeBruijn func(int, int) // recursive closure
+	ConstructDeBruijn = func(t, p int) {
+		if t > substringLength {
+			if substringLength%p == 0 {
+				seq = append(seq, a[1:p+1]...)
+			}
+		} else {
+			a[t] = a[t-p]
+			ConstructDeBruijn(t+1, p)
+			for j := int(a[t-p] + 1); j < alphabetLength; j++ {
```
TimothyStiles, line 151: What is 'j'? Same with `t, p, and a`.

TimothyStiles (changes requested):
More comments and fixing of single variable names are needed.

Should barcodes have their own file/package?

TimothyStiles:
@Koeng101 for single letter variables that you don't understand it'd be sufficient to name them something longer than a single letter as something related to what you think it does as long as reference to rosetta code function exists and you leave as much comments as possible in control flow. 

Single letter variables are literally too small a symbol to make the code visually easy to read. Just this improvement could help a ton with future debugging and help future developers figure out what's going on. Also testing will have to be a little more thorough than normal because from this perspective it's somewhat a black-ish box.

I ran into similar problems with Booth Least Rotation so for its test I took a plasmid sequence and rotated it for each letter then hashed it and checked the hash against the hash from the original rotation. Not sure what the equivalent would be here but should have a similar-ish feel.

primers.go
jkhales: In general if you can using a for loop instead of recursion is safer.

Recursion is best left for coding interviews.

primers.go
```diff
@@ -103,3 +104,108 @@ func MeltingTemp(sequence string) float64 {
...
+	var buf bytes.Buffer
+	for _, i := range seq {
+		buf.WriteByte(alphabet[i])
+	}
+	b := buf.String()
+	return b + b[0:substringLength-1] // as cyclic append first (n-1) digits
+}
+
+// CreateBarcodes creates a list of barcodes given a desired barcode length, the maxSubSequence shared in each barcode,
+// any banned sequences within the barcode, and any functions which may ban sequences within the barcode.
```
jkhales, line 167:
```suggestion
// Sequences maybe marked as banned by passing a static list, bannedSequences, or, if more flexibility is needed, through a list of `bannedFunctions` that dynamically generate bannedSequences.
// A banned sequence will do.... ?
```

Koeng101: fixed.

primers.go
```diff
@@ -103,3 +104,178 @@ func MeltingTemp(sequence string) float64 {
...
+				barcodeNum++
+			}
+		}
+		barcodes = append(barcodes, debruijn[start:end])
+	}
+	return barcodes
+}
+
+// SimpleCreateBarcodes is a simplified version of CreateBarcodes with sane defaults.
+func SimpleCreateBarcodes(length int, maxSubSequence int) []string {
```
TimothyStiles, line 279: I think this should be named `CreateBarcodes`, and the more complex version should have a different name.

TimothyStiles (changes requested):

primers.go
jkhales: `createBarcodes`
`createBarcodesWithBannedSequences`

primers.go
```diff
@@ -103,3 +104,182 @@ func MeltingTemp(sequence string) float64 {
...
+			for strings.Contains(debruijn[start:end], bannedSequence) {
+				if end+1 > len(debruijn) {
+					return barcodes
+				}
+				start++
+				end++
+				barcodeNum++
+			}
+			// Check reverse complement as well for the banned sequence
+			for strings.Contains(debruijn[start:end], ReverseComplement(bannedSequence)) {
```
TimothyStiles, line 257: This code block for checking for bannedsequences in reverse complements is untested. Test it and I think we got a merge here!

TimothyStiles (changes requested):

PR #98, opened 2021-04-22
Synthesis fixer

Koeng101:
This is the synthesis fixer. What I need to do:

- Add good tests
- Make overlap checker into a recursive function
- Add good description, delete old comments of how it works

Check it out @TimothyStiles


transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
...
+							}
+
+						}
+					}
+				}
+			}
+		}
+
+		// The following statements are the magic sauce that makes this all worthwhile.
+		// Parameters: step, gcbias, start, end, quantityfix
```
TimothyStiles, line 859: ADD SO MANY COMMENTS.

transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
...
+				case codonBias < toCodonBias:
+					db.MustExec(`INSERT INTO codonbias(fromcodon, tocodon, gcbias) VALUES (?, ?, ?)`, codon.Triplet, toCodon.Triplet, "GC")
+				}
+			}
+		}
+	}
+
+	// Insert seq and history
+	pos := 1
+	for i := 0; i < len(sequence)-3; i = i + 3 {
```
TimothyStiles, line 786: What's the +3 all about?

transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
...
+		codon := sequence[i : i+3]
+		db.MustExec(`INSERT INTO seq(pos) VALUES (?, ?)`, pos)
+		db.MustExec(`INSERT INTO history(pos, codon, step) VALUES (?, ?, 0)`, pos, codon)
+		db.MustExec(`INSERT INTO weights(pos, codon, weight) VALUES (?,?,?)`, pos, codon, weightTable[codon])
+	}
+
+	// For a maximum of 1000 iterations, see if we can do better
+	for i := 1; i < 1000; i++ {
```
TimothyStiles, line 794: Why 1000? Should this be a function variable?

transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
...
+			c <- DnaSuggestion{position, (loc[1] / 3), "NA", 1, "BsaI removal"}
+		case leftover != 0:
+			c <- DnaSuggestion{position - 1, (loc[1] / 3) - 1, "NA", 1, "BsaI removal"}
+		}
+	}
+}
+
+func findProblems(sequence string, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) []DnaSuggestion {
+	// Run functions to get suggestions
+	suggestions := make(chan DnaSuggestion, 100)
```
TimothyStiles, line 708: why 100 channels?

transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
...
+		suggestionsList = append(suggestionsList, suggestion)
+	}
+	return suggestionsList
+}
+
+func FixCds(sequence string, codontable CodonTable, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) (string, error) {
+	var db *sqlx.DB
+	db = sqlx.MustConnect("sqlite3", ":memory:")
+	createMemoryDbSql := `
+	CREATE TABLE codon (
```
TimothyStiles, line 727: ADD SO MANY COMMENTS.

transformations.go
```diff
@@ -577,3 +582,315 @@ func AddCodonTable(firstCodonTable CodonTable, secondCodonTable CodonTable) Codo
 	c.AminoAcids = finalAminoAcids
 	return c
 }
+
```
TimothyStiles, line 585: Move to a new file called `synthesis.go`

transformations.go
```diff
@@ -2,6 +2,7 @@ package poly
 
 import (
 	"errors"
+
```
jkhales, line 5: revert?

data/pichiaTable.json
```diff
@@ -0,0 +1,379 @@
+{
```
jkhales, line 1: Can you explain what the pichia table is in the PR description?

synthesis.go
```diff
@@ -0,0 +1,262 @@
+package poly
+
+import (
+	"github.com/jmoiron/sqlx"
+	//"github.com/juliangruber/go-intersect"
```
jkhales, line 5:
```suggestion
```
?

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+
+The primary goal of the synthesis fixer is to fix protein coding sequences (CDS) in preparation
+for DNA synthesis. Because CDSs are flexible in their coding, we can change codons to remove
+problematic parts of sequences. The synthesis fixer is built to be very fast while covering the
+majority of use-cases for synthesis fixing.
+
+Our synthesis fixer function (FixCds) takes in 3 parameters: a coding sequence, a codon table,
+and a list of functions that identify problematic sequences. These functions output not only
+the region where a problem occurs, but also some heuristics to help the function quickly find a
+solution to the given constraints. For example, function outputs can include a gc bias.
```
jkhales, line 27: cool 😎

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+
+The synthesis fixer works in a declarative rather than imperative approach. We spend a lot of time
+setting up our problem space so that we can quickly and easily move to a solution. First, we build
+an in-memory SQLite database with everything about our sequence - including its codon table,
+possible codon-to-codont transformations, and the sequence itself.
+
+We then run all the user-provided problematicSequenceFuncs in-parallel to generate DnaSuggestions
+of places that we should fix in our sequence. Then, we check if there are any DnaSuggestions that
+overlap - for example, perhaps a repeat region needs to be fixed in a location with a high GC content,
+so we bias fixing that repeat region to fixes that have a high AT content. If there are any overlapping
```
jkhales, line 37: just out of naive curiosity, why is having a high AT content good?

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+suggestions, those suggestions are fixed first.
+
+Once we have a list of those suggestions, we fix the sequence using some simple SQL. Once our first round
+of fixes is applied, we test the sequence again to see if we missed any problematic sequence locations. We
+do that for 1000 rounds, at which point we just give up and return the sequence to the user as is.
```
jkhales, line 42: where is this set? This sounds like it should be a variable, not hard coded.

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+
+// DnaSuggestion is a suggestion of a fixer, generated by a problematicSequenceFunc.
+type DnaSuggestion struct {
+	Start          int    `db:"start"`
+	End            int    `db:"end"`
+	Bias           string `db:"gcbias"`
+	QuantityFixes  int    `db:"quantityfixes"`
+	SuggestionType string `db:"suggestiontype"`
+	Step           int    `db:"step"`
+	Id             int    `db:"id"`
```
jkhales, line 59: What is the point of these.... comments? `db:"start"` etc.

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+				position := loc[0] / 3
+				leftover := loc[0] % 3
+				switch {
+				case leftover == 0:
+					c <- DnaSuggestion{position, (loc[1] / 3), "NA", 1, "TypeIIS removal", 0, 0}
+				case leftover != 0:
+					c <- DnaSuggestion{position, (loc[1] / 3) - 1, "NA", 1, "TypeIIS removal", 0, 0}
+				}
+			}
+		}
```
jkhales, line 98: This is the same code as `FindBsaI`.

 Make a generic `FindEnzme` function that takes it's regex and `SuggestionType` as parameters. 
 
 Then `FindBsaI` becomes:
 
 `FindEnzme(sequence string, c chan DnaSuggestion, wg *sync.WaitGroup, `GGTCTC`, "BsaI removal")`
 
 Write a unit test for this generic `FindEnzyme` function.

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+		codon := sequence[i : i+3]
+		db.MustExec(`INSERT INTO seq(pos) VALUES (?)`, pos)
+		db.MustExec(`INSERT INTO history(pos, codon, step) VALUES (?, ?, 0)`, pos, codon)
+		db.MustExec(`INSERT INTO weights(pos, codon, weight) VALUES (?,?,?)`, pos, codon, weightTable[codon])
+		pos++
+	}
+
+	var err error
+	// For a maximum of 100 iterations, see if we can do better
+	for i := 1; i < 100; i++ {
```
jkhales, line 202:
```suggestion
	for i := 1; i < ITERATION; i++ {
```
this doesn't seem like it should be hardcode.

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+		for _, suggestion := range suggestions { // if you want to add overlaps, add suggestionIndex
+			// First, let's insert the suggestions that we found using our problematicSequenceFuncs
+			_, err = db.Exec(`INSERT INTO suggestedfix(step, start, end, gcbias, quantityfixes, suggestiontype) VALUES (?, ?, ?, ?, ?, ?)`, i, suggestion.Start, suggestion.End, suggestion.Bias, suggestion.QuantityFixes, suggestion.SuggestionType)
+			if err != nil {
+				return sequence, err
+			}
+		}
+
+		// The following statements are the magic sauce that makes this all worthwhile.
+		// Parameters: step, gcbias, start, end, quantityfix
```
jkhales, line 217: nice. Can you explain a bit more why this is a good idea/likely to generate good results?

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+
+	var suggestionsList []DnaSuggestion
+	for suggestion := range suggestions {
+		suggestionsList = append(suggestionsList, suggestion)
+	}
+	return suggestionsList
+}
+
+// FixCds fixes a CDS given the CDS sequence, a codon table, and a list of functions to solve for.
+func FixCds(sqlitePath string, sequence string, codontable CodonTable, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) (string, error) {
```
jkhales, line 122:
```suggestion
func generateKoeniHeuristicSequence(sqlitePath string, sequence string, codontable CodonTable, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) (string, error) {
```
If this is a non standard generation technique, it needs more explanation. If it is a standard technique, it needs a link.

synthesis.go
```diff
@@ -0,0 +1,262 @@
...
+		        FROM   seq AS s
+		               JOIN history AS h
+		                 ON h.pos = s.pos
+		               JOIN weights AS w
+		                 ON w.pos = s.pos
+		               JOIN codon AS c
+		                 ON h.codon = c.codon
+		               JOIN codonbias AS cb
+		                 ON cb.fromcodon = c.codon
+		        WHERE `
```
jkhales, line 236: Is there nothing like python pandas for golang?

relying on sql, even if in memory, for stuff like this seems heavy, especially when computations get big.

synthesis_test.go
```diff
@@ -0,0 +1,84 @@
...
+		for _, cutSite := range []string{"GAAGAC", "GGTCTC", "GCGATG", "CGTCTC", "GCTCTTC", "CACCTGC"} {
+			if strings.Contains(optimizedSeq, cutSite) {
+				t.Errorf("phusion" + " contains " + cutSite)
+			}
+			if strings.Contains(ReverseComplement(optimizedSeq), cutSite) {
+				t.Errorf("phusion" + " reverse complement contains " + cutSite)
+			}
+		}
+	}
+}
```
jkhales, line 84: cool stuff

synthesis.go
Koeng101: Synthesizers like neutral 50:50 AT/GC content because that is easier for them to synthesize. If the region we want to fix has high GC content, then by adding AT biased fixes, we can equalize that AT/GC ratio to be nearer to 50:50.

synthesis.go
Koeng101: In almost all cases, the DNA will be fixed in 1-3 iterations. If it goes for 1000 iterations, we essentially know 100% that this thing cannot be fixed.

synthesis.go
Koeng101: I'm not really sure what you mean by a standard generation technique. This functions fixes CDSs so they can be synthesized. It's not for making heuristics or anything, it is specifically built so that users can send things to DNA synthesizers without having to worry about different sequence elements.

synthesis.go
Koeng101: I'll write more in the actual request, but the basic idea is that I am using SQL as a declarative language to choose fixes to the DNA that are most likely to work in a biological context while avoiding infinite recursion. I've attempted a version that doesn't use a declarative language, but it gets REAL hard to read really quick.

synthesis.go
Koeng101: I use SQL here as an embedded declarative language that people don't have to learn a domain specific language to use. 

Pandas wouldn't really work here, since I'm explicitly looking at doing joins from a variety of sources. The closest thing would be an embedded thing like - https://pkg.go.dev/modernc.org/ql?utm_source=godoc - but SQLite is so widely used, I thought might as well go for something where you won't have to learn an entirely new thing to use.

Koeng101: Computations stay sub-second, which is still 100x better than alternatives out there. They're pretty bounded by the size of a given gene.

synthesis.go
jkhales: cool

synthesis.go
jkhales: I see. I would include that to explain the magic number.

synthesis.go
jkhales: hm ok, will have to read your detailed explanation to understand.

synthesis.go
jkhales: > I'm not really sure what you mean by a standard generation technique

Like if I ran the sequence through a different library would you get the exact same result?

Becuase you are doing an estimate here: https://github.com/TimothyStiles/poly/pull/98/files/762d6766a445bfc287ed3e1014bc6ee854d8b961?file-filters%5B%5D=.go#diff-a7a2897e953e4b002cec604e45b7055c857a064bec21525d7b020d50538f1f4bR201
I assume not.

> It's not for making heuristics or anything
OK I see.

Wjhat guarantees do the user have about this sequence? Like is it within a certain error range? 

Or after 100 iterations are we ensured that we will get a result that is completely functional and all deviations are just noise?

synthesis.go
jkhales: sounds good. Also our convo in discord made a lot of sense.

synthesis.go
jkhales: I assume this is the function that you expected to converge after a few times, but I'd still like to know how it deviates from other implementations and if those deviations are neglible to the resutl.

Koeng101: TL;DR: no guarantees, but empirically good enough.

> Like if I ran the sequence through a different library would you get the exact same result?

Absolutely not. One thing about automated synthesis fixing is that it isn't really a popular thing right now. There is essentially 2 open source libraries that do this, and I wrote one of them (it kinda sucked). Funny enough, I actually did use Pandas in that one.

> What guarantees do the user have about this sequence? Like is it within a certain error range? Or after 100 iterations are we ensured that we will get a result that is completely functional and all deviations are just noise?

No real guarantees except I want to tell the user if we couldn't figure out a fix for their sequence.

> I'd still like to know how it deviates from other implementations

The only real competitor is https://edinburgh-genome-foundry.github.io/DnaChisel/ . The main difference is they use a constraint solving algorithm, while here we just directly fix things that we know are troublesome. 

I think the assumption you're coming with is that we know what we're looking for when it comes to synthesis fixing, but we really do not.

Do repeats of exactly 20 base pairs and no more and no less make it so you can't synthesize a given piece of DNA? Twist says they can't, Genscript says they can. Even if a synthesis provider says they can provide certain DNA molecules, they often can't (we experienced 20% failure rate with Twist with DNA they said they COULD make). When machine learning is applied to see what can or cannot be synthesized by given synthesis providers, we can approximate better results than the providers themselves. Sometimes, constraints on synthesis are also relaxed during times of excess production. 

For this implementation, I'm coming in with experience from FreeGenes of what Twist told me they can't do (though they could not do a lot more than just this). We synthesized about ~4-5 million base pairs there. As it turns out, if you do 3 things(1):

1. Remove repeats
2. Remove homopolymers
3. Remove high GC/low GC regions

While not allowing cyclic fixes (ie, same place getting flipped back and forth), you get ~5,225/5,228 genes to be successfully synthesis fixed (the other 3 you need small repeat testing). You don't need theoretically perfect methods for fixing most people's synthesis (like what DnaChisel)

(1) [Here is the implementation](https://github.com/Koeng101/dna_designer/blob/master/dna_designer/moclo.py) I used for FreeGenes. Don't judge me, it's academic code.

jkhales: Thanks for that detailed explanation. Very interesting and helpful.

I guess I would then suggest the prosaic `AttemptToFixCds` and copy some of this awesome comment into the func description.

Koeng101:
@TimothyStiles Why is code climate still yelling at me?

synthesis.go
Koeng101: Added doc saying that we only attempt to make sequences good.

Koeng101:
Before this is committed, we need the synthesis fixer to add back on the stop codons.

Koeng101:
> Before this is committed, we need the synthesis fixer to add back on the stop codons.

Fixed.

isaacguerreir (changes requested):

synthesis/synthesis.go
```diff
@@ -0,0 +1,273 @@
...
+		SELECT t.codon,
+		       t.pos,
+		       ? AS step
+		FROM   (SELECT cb.tocodon AS codon,
+		               s.pos      AS pos
+		        FROM   seq AS s
+		               JOIN history AS h
+		                 ON h.pos = s.pos
+		               JOIN weights AS w
+		                 ON w.pos = s.pos
```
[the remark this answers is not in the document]
Koeng101: > Using this line this will actually make all positions and codons with the same weight and you will not select the codons that have the highest weights and aren't the last ones used.

That line doesn't actually do any writing, so I can't really see how it could make all the positions / codons the same weight. What it should be doing is selecting all weights at a given position, which I do believe it does correctly

> Removing the pos field in the weight table, because the weights are dependent of the codon, but independent of position (you should have the same weight, for the same codon, even in different positions)

This was a specific design decision and implementation for codon harmonization purposes. I don't believe we should change this, and if it is necessary to push the PR through, we should go add codon harmonization now.

Koeng101:
@TimothyStiles CodeClimate is buggin' again.

synthesis/synthesis.go
```diff
@@ -0,0 +1,388 @@
...
+	SuggestionType string `db:"suggestiontype"`
+	Step           int    `db:"step"`
+	ID             int    `db:"id"`
+}
+
+// RemoveSequence is a generator for a problematicSequenceFuncs for specific
+// sequences.
+func RemoveSequence(sequencesToRemove []string, reason string) func(string, chan DnaSuggestion, *sync.WaitGroup) {
+	return func(sequence string, c chan DnaSuggestion, wg *sync.WaitGroup) {
+		var enzymes []string
```
isaacguerreir, line 71: It's not so important but, sequencesToRemove instead of enzymes is a better variable name than enzymes. I know, normally people will use this for removing restriction enzymes, but I think the purpose of the function is to allow users to remove more than restriction enzyme patterns.

Koeng101: Fixed.

Koeng101:
@TimothyStiles Believe this is about ready for review :)

synthesis/synthesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	ct := codon.ReadCodonJSON(dataDir + "pichiaTable.json")
+
+	var functions []func(string, chan DnaSuggestion, *sync.WaitGroup)
+	// Remove restriction enzyme sites
+	functions = append(functions, RemoveSequence([]string{"GAAGAC", "GGTCTC", "GCGATG", "CGTCTC", "GCTCTTC", "CACCTGC"}, "TypeIIS restriction enzyme site"))
```
TimothyStiles, line 28: What's with all these appends?

synthesis/synthesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func TestFixCdsWithAlteredCodonTable(t *testing.T) {
+	bla := "ATGAAAAAAAAAAGTATTCAACATTTCCGTGTCGCCCTTATTCCCTTTTTTGCGGCATTTTGCCTTCCTGTTTTTGCTCACCCAGAAACGCTGGTGAAAGTAAAAGATGCTGAAGATCAGTTGGGTGCACGAGTGGGTTACATCGAACTGGATCTCAACAGCGGTAAGATCCTTGAGAGTTTTCGCCCCGAAGAACGTTTTCCAATGATGAGCACTTTTAAAGTTCTGCTATGTGGCGCGGTATTATCCCGTATTGACGCCGGGCAAGAGCAACTCGGTCGCCGCATACACTATTCTCAGAATGACTTGGTTGAGTACTCACCAGTCACAGAAAAGCATCTTACGGATGGCATGACAGTAAGAGAATTATGCAGTGCTGCCATAACCATGAGTGATAACACTGCGGCCAACTTACTTCTGACAACGATCGGAGGACCGAAGGAGCTAACCGCTTTTTTGCACAACATGGGGGATCATGTAACTCGCCTTGATCGTTGGGAACCGGAGCTGAATGAAGCCATACCAAACGACGAGCGTGACACCACGATGCCTGTAGCAATGGCAACAACGTTGCGCAAACTATTAACTGGCGAACTACTTACTCTAGCTTCCCGGCAACAATTAATAGACTGGATGGAGGCGGATAAAGTTGCAGGACCACTTCTGCGCTCGGCCCTTCCGGCTGGCTGGTTTATTGCTGATAAATCTGGAGCCGGTGAGCGTGGGTCTCGCGGTATCATTGCAGCACTGGGGCCAGATGGTAAGCCCTCCCGTATCGTAGTTATCTACACGACGGGGAGTCAGGCAACTATGGATGAACGAAATAGACAGATCGCTGAGATAGGTGCCTCACTGATTAAGCATTGGTAA"
+
+	ct := codon.ReadCodonJSON(dataDir + "alteredPichiaTable.json")
```
TimothyStiles, line 52: No abbreviations. Call it a codonTable.

synthesis/synthesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+	phusion := "MGHHHHHHHHHHSSGILDVDYITEEGKPVIRLFKKENGKFKIEHDRTFRPYIYALLRDDSKIEEVKKITGERHGKIVRIVDVEKVEKKFLGKPITVWKLYLEHPQDVPTIREKVREHPAVVDIFEYDIPFAKRYLIDKGLIPMEGEEELKILAFDIETLYHEGEEFGKGPIIMISYADENEAKVITWKNIDLPYVEVVSSEREMIKRFLRIIREKDPDIIVTYNGDSFDFPYLAKRAEKLGIKLTIGRDGSEPKMQRIGDMTAVEVKGRIHFDLYHVITRTINLPTYTLEAVYEAIFGKPKEKVYADEIAKAWESGENLERVAKYSMEDAKATYELGKEFLPMEIQLSRLVGQPLWDVSRSSTGNLVEWFLLRKAYERNEVAPNKPSEEEYQRRLRESYTGGFVKEPEKGLWENIVYLDFRALYPSIIITHNVSPDTLNLEGCKNYDIAPQVGHKFCKDIPGFIPSLLGHLLEERQKIKTKMKETQDPIEKILLDYRQKAIKLLANSFYGYYGYAKARWYCKECAESVTAWGRKYIELVWKELEEKFGFKVLYIDTDGLYATIPGGESEEIKKKALEFVKYINSKLPGLLELEYEGFYKRGFFVTKKRYAVIDEEGKVITRGLEIVRRDWSEIAKETQARVLETILKHGDVEEAVRIVKEVIQKLANYEIPPEKLAIYEQITRPLHEYKAIGPHVAVAKKLAAKGVKIKPGMVIGYIVLRGDGPISNRAILAEEYDPKKHKYDAEYYIENQVLPAVLRILEGFGYRKEDLRYQKTRQVGLTSWLNIKKSGTGGGGATVKFKYKGEEKEVDISKIKKVWRVGKMISFTYDEGGGKTGRGAVSEKDAPKELLQMLEKQKK*"
+	ct := codon.ReadCodonJSON(dataDir + "pichiaTable.json")
+	var functions []func(string, chan DnaSuggestion, *sync.WaitGroup)
+	functions = append(functions, RemoveSequence([]string{"GAAGAC", "GGTCTC", "GCGATG", "CGTCTC", "GCTCTTC", "CACCTGC"}, "TypeIIS restriction enzyme site."))
+	for i := 0; i < b.N; i++ {
```
TimothyStiles, line 70: what is `i`?

synthesis/synthesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+	shouldChangeTo := "Changed position 9 from CGT to CGA for reason: Should change to CGA with the Altered Picha Table, because I choose this to be highest. Complete sequence: ATGAAAAAAAAAAGTATTCAACATTTCCGAGTCGCCCTTATTCCCTTTTTTGCGGCATTTTGCCTTCCTGTTTTTGCTCACCCAGAAACGCTGGTGAAAGTAAAAGATGCTGAAGATCAGTTGGGTGCACGAGTGGGTTACATCGAACTGGATCTCAACAGCGGTAAGATCCTTGAGAGTTTTCGCCCCGAAGAACGTTTTCCAATGATGAGCACTTTTAAAGTTCTGCTATGTGGCGCGGTATTATCCCGTATTGACGCCGGGCAAGAGCAACTCGGTCGCCGCATACACTATTCTCAGAATGACTTGGTTGAGTACTCACCAGTCACAGAAAAGCATCTTACGGATGGCATGACAGTAAGAGAATTATGCAGTGCTGCCATAACCATGAGTGATAACACTGCGGCCAACTTACTTCTGACAACGATCGGAGGACCGAAGGAGCTAACCGCTTTTTTGCACAACATGGGGGATCATGTAACTCGCCTTGATCGTTGGGAACCGGAGCTGAATGAAGCCATACCAAACGACGAGCGTGACACCACGATGCCTGTAGCAATGGCAACAACGTTGCGCAAACTATTAACTGGCGAACTACTTACTCTAGCTTCCCGGCAACAATTAATAGACTGGATGGAGGCGGATAAAGTTGCAGGACCACTTCTGCGCTCGGCCCTTCCGGCTGGCTGGTTTATTGCTGATAAATCTGGAGCCGGTGAGCGTGGGTCTCGCGGTATCATTGCAGCACTGGGGCCAGATGGTAAGCCCTCCCGTATCGTAGTTATCTATACGACGGGGAGTCAGGCAACTATGGATGAACGAAATAGACAGATCGCTGAGATAGGTGCCTCACTGATTAAGCATTGGTAA"
+	if textChange != shouldChangeTo {
+		t.Errorf("%s\nshould be\n%s", textChange, shouldChangeTo)
+	}
+}
+
+func BenchmarkFixCds(b *testing.B) {
```
TimothyStiles, line 65: How does this benchmark work?

synthesis/synthesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+	var functions []func(string, chan DnaSuggestion, *sync.WaitGroup)
+	functions = append(functions, RemoveSequence([]string{"GGTCTC", "CGTCTC"}, "TypeIIS restriction enzyme site."))
+	_, _, err := FixCds(":memory:", seq, ct, functions)
+	if err != nil {
+		t.Errorf("Failed with error: %s", err)
+	}
+}
+
+func TestFixCds(t *testing.T) {
+	ct := codon.ReadCodonJSON(dataDir + "pichiaTable.json")
```
TimothyStiles, line 103: No abbreviations.

TimothyStiles (changes requested):
There's more to review when it comes to code coverage in tangentially modified packages outside of the new synthethesis sub-package but something seems to be breaking my IDE test build. Will fix and add more but everything I commented on can be fixed before that @koeng101.

Koeng101:
Question @TimothyStiles : What if I can't figure out how to test some of the errors? The easiest thing to do is to delete the error checking to improve coverage, but there might be an edge case that gets complicated without it.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/TimothyStiles/poly/checks"
+	"github.com/TimothyStiles/poly/transform"
+	"github.com/TimothyStiles/poly/transform/codon"
+	"github.com/jmoiron/sqlx"
+	_ "modernc.org/sqlite" // imports CGO-less sqlite
+)
+
+// FixIterations is the standard default for the number of rounds FixCds will
+// iterate for while trying to find a fixed solution for a sequence.
+var FixIterations = 100
```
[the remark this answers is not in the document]
Koeng101: Fair, seems like a reasonable function argument. Will fix this.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+	QuantityFixes  int    `db:"quantityfixes"`
+	SuggestionType string `db:"suggestiontype"`
+	Step           int    `db:"step"`
+	ID             int    `db:"id"`
+}
+
+// RemoveSequence is a generator for a problematicSequenceFuncs for specific
+// sequences.
+func RemoveSequence(sequencesToRemove []string, reason string) func(string, chan DnaSuggestion, *sync.WaitGroup) {
+	return func(sequence string, c chan DnaSuggestion, wg *sync.WaitGroup) {
```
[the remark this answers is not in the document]
Koeng101: I'm not sure if this should be explained in code or not. 

Basically, the generator lets you add information to the function to have it work better. For example, you might want to add in a fixing function that takes into account the genomic sequence of an organism to remove kmer matches. That function will need a genomic sequence as an input. But what if you want to also remove BsaI sites? Well, the strong typing of Golang will make it so those two functions have to have the same inputs, even if they require slightly different kinds of data. 

So, we generate a standard kind of function (for typing reasons) for each different kind of thing we want to check for.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+				}
+			}
+		}
+		wg.Done()
+	}
+}
+
+// RemoveRepeat is a generator to make a problematicSequenceFunc for repeats.
+func RemoveRepeat(repeatLen int) func(string, chan DnaSuggestion, *sync.WaitGroup) {
+	return func(sequence string, c chan DnaSuggestion, wg *sync.WaitGroup) {
```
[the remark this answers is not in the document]
Koeng101: See above.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		wg.Done()
+	}
+
+}
+
+func findProblems(sequence string, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) []DnaSuggestion {
+	// Run functions to get suggestions
+	suggestions := make(chan DnaSuggestion, 100)
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
Koeng101: https://gobyexample.com/waitgroups

https://golang.org/src/sync/waitgroup.go

https://pkg.go.dev/sync#example-WaitGroup

Pretty sure `wg` is idiomatic Golang code.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+			     suggestedfix)
+		SELECT t.codon,
+		       t.pos,
+		       ? AS step,
+		       ? AS suggestedfix
+		FROM   (SELECT cb.tocodon AS codon,
+		               s.pos      AS pos
+		        FROM   seq AS s
+		               JOIN history AS h
+		                 ON h.pos = s.pos
```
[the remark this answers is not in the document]
Koeng101: history

(`history AS h`)

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+		               JOIN codon AS c
+		                 ON h.codon = c.codon
+		               JOIN codonbias AS cb
```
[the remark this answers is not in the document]
Koeng101: Yaknow, you might be right there. This is just how they always do it in SQL, not entirely sure why. I'll test out alternatives.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+synthesis fixer takes advantage of synonymous codons in protein coding
+sequences (CDS) to remove problematic sequences that either users don't want
+(like restriction enzymes sites) or that would cause DNA synthesis companies to
+reject a synthesis project.
+
+This synthesis fixer is meant to cover the majority of use cases for DNA
+fixing. It is not intended to cover all possible use cases, since the majority
+of DNA design does not actually have these edge cases.
+
+FixCds does not guarantee that all requested features will be removed. If you
```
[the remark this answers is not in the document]
Koeng101: > takes advantage of synonymous codons in protein coding sequences (CDS)

Do you want me to explain what protein coding sequence means or something? Little confused there.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+		for i := 0; i < len(sequence)-repeatLen; i++ {
+			_, alreadyFoundFor := kmers[sequence[i:i+repeatLen]]
+			_, alreadyFoundRev := kmers[transform.ReverseComplement(sequence[i:i+repeatLen])]
+			kmers[sequence[i:i+repeatLen]] = true
+			if alreadyFoundFor || alreadyFoundRev {
+				position := i / 3
+				leftover := i % 3
+				switch {
+				case leftover == 0:
+					c <- DnaSuggestion{position, ((i + repeatLen) / 3), "NA", 1, "Repeat sequence"}
```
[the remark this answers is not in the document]
Koeng101: Added docs about what NA means (Bias can be GC, AT, or NA)

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+	close(suggestions)
+
+	var suggestionsList []DnaSuggestion
+	for suggestion := range suggestions {
+		suggestionsList = append(suggestionsList, suggestion)
+	}
+	return suggestionsList
+}
+
+// FixCds fixes a CDS given the CDS sequence, a codon table, and a list of functions to solve for.
```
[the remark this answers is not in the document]
Koeng101: Well, presumably they import FixCDS in the `synthesis` package, so synthesis.FixCDS means they're fixing the CDS for synthesis. Is that not good enough?

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+func FixCds(sqlitePath string, sequence string, codontable codon.Table, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup)) (string, []Change, error) {
+
+	if len(sequence)%3 != 0 {
+		return "", []Change{}, errors.New("this sequence isn't a complete CDS, please try to use a CDS without interrupted codons")
+	}
+
+	db := sqlx.MustConnect("sqlite", sqlitePath)
```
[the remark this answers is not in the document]
Koeng101: For this particular little section of SQL, that kinda seems like putting on a comment to explain `fmt.Println("Hello World")`. Do you perhaps mean explain to explain the schema?

synthesis/synthesis_test.go
Koeng101: Well, you gotta input a list of functions that make suggestions. I removed the appends and just made a big ole list in the function itself.

synthesis/synthesis_test.go
Koeng101: What do you mean? It runs a Golang benchmark. It's like a test, except a benchmark.

synthesis/synthesis.go
```diff
@@ -0,0 +1,336 @@
...
+		sequence = strings.Join(codons, "")
+	}
+
+	var changes []Change
+	_ = db.Select(&changes, `SELECT h.pos AS position, h.step AS step, (SELECT codon FROM history WHERE pos = h.pos AND step = h.step-1 LIMIT 1) AS codonfrom, h.codon AS codonto, sf.suggestiontype AS reason FROM history AS h JOIN suggestedfix AS sf ON sf.id = h.suggestedfix WHERE h.suggestedfix IS NOT NULL ORDER BY h.step, h.pos`)
+
+	return sequence, changes, errors.New("Could not find a solution to sequence space")
+}
+
+// FixCdsSimple is FixCds with some defaults for normal usage, including
```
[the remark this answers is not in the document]
Koeng101: It literally says what it does right there

synthesis/synthesis.go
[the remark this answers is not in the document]
Koeng101: https://www.reddit.com/r/SQL/comments/hfrjs9/are_single_letter_table_aliases_still_the_most/

It's generally accepted practice, and after modifying a few of them to be more verbose, I think it hurts readability. Unlike using single letter variables in normal code, in SQL you have a clearly defined alias right nearby within the query.

Issue #92, opened 2021-04-13
all_iupac_variants port from easy_dna


tijeco:
From [easy_dna](https://edinburgh-genome-foundry.github.io/easy_dna/). Return all unambiguous possible versions of the given sequence.

```python
>>> all_iupac_variants('ATN')
>>> ['ATA', 'ATC', 'ATG', 'ATT'] 
```

Will be using the ambiguity codes from [here](https://genome.ucsc.edu/goldenPath/help/iupac.html).

TimothyStiles:
@tijeco any recent progress?

tijeco:
@TimothyStiles thanks for checking in! I have working on this on my calendar for tomorrow!

tijeco:
I got a chance to work on this today!

I thought I had a solution until I did some tests.... Will keep at it though! 

The original implementation had a quippy one liner with itertools, so I'm thinking some sort of recursion might be needed.

It's a fun problem!

PR #106, opened 2021-04-23
All iupac variants

tijeco:
This is for #92, a port of all_iupac_variants from easy_dna.

IUPAC codes are stored as a map of rune slices. The cartesian products are then generated using a retooled version of the function found in github.com/schwarmco/go-cartesian-product, which uses recursion and go routines, so it should scale quite nicely to large sequences.


sequence.go
```diff
@@ -92,3 +93,75 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+		// inc WaitGroup
+		wg.Add(1)
+		// create copy of result
+		resultCopy := append([]rune{}, result...)
+		// call self with remaining params
+		go iterate(wg, channel, append(resultCopy, p[i]), params...)
+	}
+}
+
+func allVariantsIUPAC(seq string) []string {
```
TimothyStiles, line 136: Make public (capitalize `allVariantsIUPAC` -> `AllVariantsIUPAC` ) and create export comment starting as `// AllVariantsIUPAC (insert description here)`

sequence.go
```diff
@@ -92,3 +93,75 @@ func getFeatureSequence(feature Feature, location Location) string {
 
 	return sequenceString
 }
+
+// the following functions Iter and iter, derive from github.com/schwarmco/go-cartesian-product
+// which uses interfaces, so I modified it to use runes
+func Iter(params ...[]rune) chan []rune {
```
TimothyStiles, line 99: make private.

TimothyStiles:
This is awesome @tijeco! I've added a couple things and will review more later today when I get the chance.

Commenting looks good. Variables could have more descriptive names. Are there any standard test cases we could use to test it?

tijeco:
@TimothyStiles thanks! From easy_dna the sample was as follows

```python
>>> all_iupac_variants​('ATN')
>>>​ [​'ATA'​, ​'ATC'​, ​'ATG'​, ​'ATT']
```

The output from the port I wrote won't be sorted, and I think because it uses go routines the order could be different each time. 

So a test would either need to sort the output, or the function would need to return sorted output. 

I don't know which is preferred.

TimothyStiles:
In this case I think we can just sort the output for the test. There may be something we can do about it be returned from the function sorted but we don't have to do that right now.

Do you think we could find a stronger test case? Something like "The quick brown fox jumped over the lazy dog", but for degenerate base pairs? N, B, V, etc?

tijeco:
@TimothyStiles Sounds good! I will work on a test set and put it in a separate pull request for sequence_test.go

TimothyStiles:
@tijeco could you put it in this pull request? Would be easier for me to manage. All you have to do is push commits to the PR in the branch and it should auto update.

tijeco:
@TimothyStiles  Will do! Thanks!

I don't know if I'll have time to work on it today/this weekend, but I definitely will early next week.

Koeng101:
So here is a potential problem: RAM issues.

Each goroutine takes about 2Kb of memory https://stackoverflow.com/questions/22326765/go-memory-consumption-with-many-goroutines

If you have 10N, for example, this would spawn (potentially) 1,048,576 goroutines, taking up nearly 2GB of memory. Meanwhile, if you did an iterative function without goroutines, you could possibly reduce that amount of RAM by a large amount because a new goroutine wouldn't be spawned for every sequence branch. Since append functions do not take much CPU, you might get equivalent performance with a much simpler stack that takes far less RAM.

Could you try some harder benchmarks to see where the concurrent method breaks down?

tijeco:
@Koeng101 thanks for pointing that out! I'll be honest, go routines and concurrency is still a new thing to me, so I'm glad to have this opportunity to work on this. 

I'll try to make a non-concurrent version of the function as well and do some benchmarks to see how memory/cpu usage varies between the two.

sequence.go
tijeco: @TimothyStiles Does the export comment go on the same line as the function declaration?

TimothyStiles:
@tijeco can you pull down the latest from `prime` and merge it with your PR branch? Otherwise it'll delete some recent changes to the file. After that I should be able to run CI and do a proper review.

tijeco:
@TimothyStiles I think I did the thing?

sequence.go
```diff
@@ -92,3 +92,77 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+		}
+	}
+	return allVariants
+}
+
+func AllVariantsIUPAC(seq string) []string {
+	//AllVariantsIUPAC( seq string) takes a string as input
+	// and returns all iupac variants as output
+	seqVariantList := [][]rune{}
+	seqVariants := []string{}
```
Koeng101, line 138: If there is only the option for uppercase runes, the sequence should be set to uppercase before it is used in the seqVariantList.

sequence.go
```diff
@@ -92,3 +92,77 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+	var possibleVariants int = 1 // a counter used to determine the possible number of variants
+	for _, inList := range inList {
+		possibleVariants *= len(inList)
+	}
+	if possibleVariants == 0 {
+		return nil // in the future this could be part of an error return?
+	}
+	allVariants := make([][]rune, possibleVariants)              // this is the 2D slice where all variants will be stored
+	variantHolders := make([]rune, possibleVariants*len(inList)) // this is an empty slice with a length totaling the size of all input characters
+	variantChoices := make([]int, len(inList))                   // these will be all the possible variants
```
Koeng101, line 110: Why are these allocations necessary? Can you run them without?

sequence.go
```diff
@@ -92,3 +92,77 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+
+		for variantChoicesIndex, variantChoice := range variantChoices {
+			variantHolder[variantChoicesIndex] = inList[variantChoicesIndex][variantChoice]
+		}
+		for variantChoicesIndex := len(variantChoices) - 1; variantChoicesIndex >= 0; variantChoicesIndex-- {
+			variantChoices[variantChoicesIndex]++
+			if variantChoices[variantChoicesIndex] < len(inList[variantChoicesIndex]) {
+				break
+			}
+			variantChoices[variantChoicesIndex] = 0
```
Koeng101, line 128: Is there a way to simplify this block of code?

sequence.go
tijeco: These allocations are helping to reduce the total memory footprint. Full disclosure: this is a retool of the cartesian product from Rosetta code https://rosettacode.org/wiki/Cartesian_product_of_two_or_more_lists

TimothyStiles:
@tijeco I did some finagling and fixed some bugs and the initial test itself. Think you could add an example function that will be rendered to our docs similar to [this](https://github.com/TimothyStiles/poly/blob/907e206826db02295287ea0da3961325886aad95/sequence_test.go#L75)?

sequence.go
tijeco: This has now been resolved!

sequence.go
tijeco: Not that I'm currently aware of

PR #107, opened 2021-04-24
Adds Rhea Parser to Poly

### LINKED ISSUE #99: Rhea parser integration
Integration and testing of the Rhea parser package into Poly

https://github.com/Koeng101/rhea

Integration and testing of the Rhea parser package into Poly

@TimothyStiles  writing a Rhea parser from scratch

**Additional context**
The RDF XML Rhea database drop can be downloaded here - https://www.rhea-db.org/help/download

Koeng101:
Closes #99 

This pull request adds a sub directory of `parsers/` to Poly with the first parser of the series, a Rhea parser for the Rhea reactions database (https://www.rhea-db.org/). 

It enables users to use a simplified higher-level struct than the raw XML provided by Rhea. It also allows users to put all that information in a normalized SQLite database. This normalized SQLite database, in-memory, also lets us test that all of our parsing of Rhea is maintaining its proper relationships. 

Though only 80% tested, this is largely because of error handling during SQL insertion, which is a bit of a hard problem, since I'd have to generate fake purposely-bad Rhea databases for every circumstance. I could also remove em and only have the final commit be checked, which wouldn't give nearly as much information, but could still work.

Other than the sqlite3 database driver, the Rhea parser uses no library outside of the standard library. 

Would appreciate a code review @TimothyStiles


Koeng101:
Shoot, I accidentally checked out of the synthesis fixer. Gonna go back and fix...

PR #109, opened 2021-04-24
Add anonymized_record function to sequence.go

isaacguerreir:
Hey, Tim

I make the anonymized_record function. Basically it's try to anonymized information of the sequence object and features. I re-implement this function from easy_dna. This is the permalink: [anonymized_record from easy_dna)](https://github.com/Edinburgh-Genome-Foundry/easy_dna/blob/14f694d549682bd0e2aeaa136cc82c0f20015c11/easy_dna/record_operations.py#L89).

[Here](https://github.com/Edinburgh-Genome-Foundry/easy_dna/blob/master/tests/test_record_operations.py) you can see the tests in easy_dna for this functionality.


sequence.go
```diff
@@ -92,3 +92,20 @@ func getFeatureSequence(feature Feature, location Location) string {
 
 	return sequenceString
 }
+
+func anonymized_record(sequence Sequence, sequence_id string) Sequence {
```
TimothyStiles, line 96: Rename AnonymizeRecord.

TimothyStiles: Also add docstring comment above:

``//AnonymizeSequence takes a sequence struct and deletes all meta data and returns a Sequence struct with just the sequence string.``

sequence_test.go
```diff
@@ -69,3 +69,26 @@ func TestLocationParser(t *testing.T) {
 		t.Errorf("Feature sequence parser has changed on test 'complement(join(893..1098,1101..2770))'. Got this:\n%s instead of \n%s", featureComplementJoin, seqComplementJoin)
 	}
 }
+
+func TestAnonymizedRecord(t *testing.T) {
```
TimothyStiles, line 73: Rename: `TestAnonymizeSequence`

sequence_test.go
```diff
@@ -69,3 +69,26 @@ func TestLocationParser(t *testing.T) {
...
+
+    for i, Feature := range anonymousSequence.Features {
+        if (sequence.Features[i].Sequence != anonymousSequence.Features[i].Sequence) {
+                t.Errorf("Should return for each Feature the same sequence. Got this: \n%s instead of \n%s", sequence.Features[i].Sequence, anonymousSequence.Features[i].Sequence)
+        }
+        if (Feature.Name != "anonymized") {
+                t.Errorf("Should return from each Feature.id in the Sequence as anonymized. Got this: \n%s instead of \n%s", Feature.Name, "anonymized")
+        }
+    }
+}
```
TimothyStiles, line 94: Add `ExampleAnonymizeSequence`

sequence.go
```diff
@@ -92,3 +92,21 @@ func getFeatureSequence(feature Feature, location Location) string {
 
 	return sequenceString
 }
+
+//AnonymizeSequence takes a sequence struct and substitute identification data for a new anonymized string and returns this modified Sequence struct.
+func anonymizeSequence(sequence Sequence, sequence_id string) Sequence {
```
TimothyStiles, line 97: capitalize first letter.

sequence_test.go
```diff
@@ -69,3 +70,37 @@ func TestLocationParser(t *testing.T) {
 		t.Errorf("Feature sequence parser has changed on test 'complement(join(893..1098,1101..2770))'. Got this:\n%s instead of \n%s", featureComplementJoin, seqComplementJoin)
 	}
 }
+
+func ExampleAnonymizeSequence(t *testing.T) {
```
TimothyStiles, line 74: remove `t *testing.T ` argument declaration. Isn't used in this kind of test and is breaking test run.

sequence.go
```diff
@@ -92,3 +92,21 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+
+        anonymizer := "anonymized"
```
TimothyStiles, line 99: this parts actually unneeded ID is generated via hashing and is directly related to sequence string itself anyways.

sequence.go
```diff
@@ -92,3 +92,21 @@ func getFeatureSequence(feature Feature, location Location) string {
...
+
+        if (sequence_id != "") {
+            anonymizer = sequence_id
+        }
+
+        sequence.Meta.Name = anonymizer
+        sequence.Meta.Keywords = ""
+        for i := range sequence.Features {
```
TimothyStiles, line 107: Is this how easy_dna does it? I thought they would have just created a new record and inserted ID + sequence string.

isaacguerreir:
Hey, Tim.

I make some commits to correct this problem. Let me know if I need to correct more things for this PR.

sequence.go
isaacguerreir: They [create a deepcopy](https://github.com/Edinburgh-Genome-Foundry/easy_dna/blob/14f694d549682bd0e2aeaa136cc82c0f20015c11/easy_dna/record_operations.py#L106) of the sequence and after that [assigns the ID for the anonymous value](https://github.com/Edinburgh-Genome-Foundry/easy_dna/blob/14f694d549682bd0e2aeaa136cc82c0f20015c11/easy_dna/record_operations.py#L112). 

Do you think deepcopy it's a necessary step for this?

** The links are permalinks for the related lines of code in easy_dna **

sequence.go
isaacguerreir: Perfect. You think have an underscore in the params as sequence_id is a problem also?

TimothyStiles: Yes. Golang style convention is camel case for almost everything except exported function.

TimothyStiles:
@isaacguerreir and I discovered that the reference implementation he was using didn't actually anonymize records and now we're just confused as to what this was supposed to be. Definitely a miscommunication on my end. Closing for now but could open back up in future.

PR #111, opened 2021-04-25
Rebase parser

### LINKED ISSUE #101: NEB REBASE parser
It would be great to have a full suite of enzymes available for `clone.go`. The easiest way to do this would be to make a parser for one of the data dumps from NEB REBASE http://rebase.neb.com/rebase/rebase.serv.html

The easiest format will probably be format #31 http://rebase.neb.com/rebase/rebase.f31.html

getBaseRestrictionEnzyme should have all restriction enzymes available with it, built into poly.

N/A

**Additional context**
REBASE is awesome and can provide customized data dumps. Though that probably shouldn't be necessary, it is something to think about.

Koeng101:
Closes #101 

Added a Rebase parser. We still need a way to convert from the enzyme cut sites to the format used by the cloning simulator, but that will come in another PR.


TimothyStiles:
@Koeng101 Things before merge:

* I'm not sure how we want to deploy this largish file. Would rather not add a tracked 4MB file into `prime`. Thoughts? Git lfs or something?

* I need to understand how this will affect project docs etc. Even if merged and released there may have to be unmerging and releasing just to test stuff out with godocs before final merge.

* Can we change `link_withrefm.txt` to something more readable?

Koeng101:
> Code climate is saying there's no issues but also that it doesn't give inline comments for PRs with files this big. Not sure if no issues means that it didn't analyze it at all or what.

Guess gotta do a manual check then... I think the code is pretty good.

> I'm not sure how we want to deploy this largish file. Would rather not add a tracked 4MB file into prime. Thoughts? Git lfs or something?

If it's only once it might be ok. Not like a monthly thing. No other thoughts really, since you need it for testing. 

> I need to understand how this will affect project docs etc. Even if merged and released there may have to be unmerging and releasing just to test stuff out with godocs before final merge.

Ok

> Can we change link_withrefm.txt to something more readable?

I don't think this would be a good idea, since that is literally the name of the file when you download it from Rebase - keeping the name the same as what someone would download it as is a good idea IMO. That's also commented in the large comment^

TimothyStiles:
@Koeng101 what's the status on the smaller rebase file? Any thoughts on deployment since it would actually need to be shipped with each release?

parsers/rebase/rebase.go
```diff
@@ -0,0 +1,264 @@
...
+	RecognitionSequence    string   `json:"recognitionSequence"`
+	MethylationSite        string   `json:"methylationSite"`
+	MicroOrganism          string   `json:"microorganism"`
+	Source                 string   `json:"source"`
+	CommercialAvailability []string `json:"commercialAvailability"`
+	References             string   `json:"references"`
+}
+
+// ParseRebase parses the Rebase database into a map of enzymes
+func ParseRebase(file []byte) map[string]Enzyme {
```
TimothyStiles, line 170: Should this be exposed to end users? It seems like this would be better as a private function for Poly developers to use unless NEB asks us to export it for them to use.

parsers/rebase/rebase.go
```diff
@@ -0,0 +1,264 @@
...
+			// After every <8> a new enzyme will start. So here, we put the current enzyme into the enzymeMap and setup a new enzyme to be filled
+			enzymeMap[enzyme.Name] = enzyme
+			enzyme = Enzyme{}
+		}
+	}
+	return enzymeMap
+}
+
+// ReadRebase returns an enzymeMap from a Rebase data dump
+func ReadRebase(path string) (map[string]Enzyme, error) {
```
TimothyStiles, line 248: `ReadRebase` makes sense to developers but should we use a name like `GetEnzymeMap` for those who don't know what rebase is?

TimothyStiles (commented):
Code looks solid. Going to check coverage in a second.

My inline comments revolve around naming and exporting. Essentially. How do we pose this to users and devs?

parsers/rebase/rebase.go
Koeng101: Should we have a function GetEnzymeMap that downloads the file directly from NEB and parses? Seems like we'd be hitting their endpoints quite hard in that case.

Koeng101: Since it would run on every test.

parsers/rebase/rebase.go
Koeng101: I was thinking of having it exposed for end users in case they want to store the enzymes elsewhere, in which they can just pass bytes in to the function. Kinda similar to the idea of `GetEnzymeMap`, where you actually just directly download the file and pass it into ParseRebase.

PR #113, opened 2021-04-25
Added Fasta parser

Koeng101:
Added Fasta subpackage parser. I moved the stuff from `io.go` and `io_test.go` to the parsers directory for a specific fasta parser and merged my concurrent fasta parser from #83 + Jordan Campbell fasta functions from #37 so that we don't have two alternative ways to parse fasta files.

Includes 100% test coverage, and all the tests and goodness from io.go.


Koeng101:
Fasta fails due to the CLI. Lemme figure this one out @TimothyStiles

parsers/fasta/fasta.go
```diff
@@ -0,0 +1,165 @@
...
+	"os"
+	"strings"
+)
+
+/******************************************************************************
+Apr 25, 2021
+
+FASTA Parser begins here
+
+Many thanks to Jordan Campbell (https://github.com/0x106) for building the first
```
TimothyStiles, line 18: I heavily refactored Jordan's stuff a while back so I guess I'm also responsible for this.

parsers/fasta/fasta.go
```diff
@@ -0,0 +1,165 @@
...
+
+	var outputFastas []Fasta
+	for fasta := range fastas {
+		outputFastas = append(outputFastas, fasta)
+	}
+	return outputFastas
+}
+
+// ParseFASTAConcurrent concurrently parses a given Fasta file in an io.Reader into a channel of Fasta structs.
+func ParseFASTAConcurrent(r io.Reader, sequences chan<- Fasta) {
```
TimothyStiles, line 63: The last parser was technically a multi-fasta parser that would add all Fastas into one struct. It would also annotate each segment based on its comment. Does this do that too?

parsers/fasta/fasta.go
```diff
@@ -0,0 +1,165 @@
...
+	close(sequences)
+}
+
+/******************************************************************************
+
+Start of FASTA Read functions
+
+******************************************************************************/
+
+// ReadFASTAGz concurrently reads a gzipped Fasta file into a Fasta channel.
```
TimothyStiles, line 115: Borked comment name needs to be ReadFASTAGzConcurrent

parsers/fasta/fasta.go
TimothyStiles: It looks like it doesn't interface with the Sequence struct at all which makes a lot of methods moot. Can this be fixed?

TimothyStiles (dismissed):

parsers/fasta/fasta.go
Koeng101: > The last parser was technically a multi-fasta parser that would add all Fastas into one struct. It would also annotate each segment based on its comment. Does this do that too?

No, and it can't. That method rapidly inflated the byte size of the sequence struct, and we can't parse a 400Gb FASTA file because the memory requirement in the design of annotating each segment into a single sequence, meaning we'd have to break it up.

> It looks like it doesn't interface with the Sequence struct at all which makes a lot of methods moot. Can this be fixed?

No. For simplicity sake, FASTA files ARE different than Genbank files in non-trivial ways (for example, one has type information for seqhash, one does not). They should be able to share an interface, but I don't think they have or should have equivalent methods.

Koeng101:
Oh crap, I have to go fix the CLI.

TimothyStiles:
> Oh crap, I have to go fix the CLI.

looool That dang CLI man. So glad we're only supporting file conversion and hashing going forward. We may want to even just move that code to a new repo.

TimothyStiles:
> Oh crap, I have to go fix the CLI.

looool That dang CLI @Koeng101. So glad we're only supporting file conversion and hashing going forward. We may want to even just move that code to a new repo.

TimothyStiles:
@Koeng101 got time to fix this and merge it with me?

TimothyStiles (approved):

PR #132, opened 2021-05-03
random_dna_sequence preliminary Go code for Easy DNA

maplemuffet:
Hello! I am not super familiar with Go, so there are a couple of changes I wasn't sure if I made correctly (for instance, dictionaries). Could you take a look and let me know how to modify this code?


TimothyStiles:
It looks like multiple things are in this PR besides the 'RandomDNA' code. @NickNolan did some of your primer stuff get wrapped in here somehow?

PR #138, opened 2021-05-05
testing out new pr coverage upload scheme.

TimothyStiles:
PR coverage is borked on forks and this fixes that by securely generating report in one context and then passing report to less secure context with secrets to be uploaded to code climate.


TimothyStiles:
Normally I wouldn't merge before confirming that this works but it seems that in order to test `workflow-run` triggers the file containing the trigger has to be merged into the default branch. Ughhh.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

PR #147, opened 2021-05-12
refactor fileParser

jkhales:
# why

The logic of `fileParser` is a bit weird.

`poly convert -i gbk file.gff` will parse correctly, but `poly convert -i gff file.gbk` will error. (assuming the file extensions are orrect).

This PR assumes the `-i` is an input type override: https://github.com/TimothyStiles/poly/blob/prime/poly/main.go#L98
 and cleans up the logic of `fileParser`. 

`fileParser` is then the same function as `fileExt`, except that it allows the user to "override" the file format. 
 Since the override is is ignored if it is the empty string `""`, `fileParser( *, "") == fileExt(*)`  and we can get rid of `fileExt`.

# test
* added an test for fileParser and a mislabelled JSON file to test its override.
* since fileExt was previously being used only as a test util, it's replacement by `fileParser(*,"")` increases the test coverage.


poly/commands.go
```diff
@@ -77,7 +78,14 @@ func convertCommand(c *cli.Context) error {
 
 		// TODO write basic check to see if input flag or all paths have accepted file extensions.
 
-		// TODO write basic check for reduduncy. I.E converting gff to gff, etc.
+		// TODO write basic check for redundancy. I.E converting gff to gff, etc.
```
jkhales, line 81: CI is blaming me for this new TODO, but I am innocent! Only fixed a typo.

jkhales:
Found and mistake, redoing.

PR #148, opened 2021-05-12
Delete file parser

jkhales:
# why

Thinking about this: `// TODO write basic check to see if input flag or all paths have accepted file`

Noticed, the weird logic of `fileParser` is probably the reason behind this TODO:

`poly convert -i gbk file.gff` will parse correctly, but `poly convert -i gff file.gbk` will error. (assuming the file extensions are correct).

This PR proposes that the `-i` should only be used when the file is being piped: `cat data/bsub.gbk | poly c -i gbk -o json > test.json`

# how

delete `fileParse`, replace calls to it with `parseExt`.

This has the added benefit of increasing test coverage, since `fileParse` is not in any tests, while `parseExt` in only in tests. 

# question

how should passing `-i` be handled when the input source is not a pipe? 

Currently it is ignored, but probably should throw an error. Unsure how much this repo follows the "fail fast and loudly" philosophy though.

### DIFF
--- a/poly/commands.go
+++ b/poly/commands.go
@@ -75,8 +75,6 @@ func convertCommand(c *cli.Context) error {
 		// gets glob pattern matches to determine which files to use.
 		matches := getMatches(c)
 
-		// TODO write basic check to see if input flag or all paths have accepted file extensions.
-
 		// TODO write basic check for reduduncy. I.E converting gff to gff, etc.
 
 		// declaring wait group outside loop
--- a/poly/commands.go
+++ b/poly/commands.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -33,6 +34,8 @@ TTFN,
 Tim
 
 ******************************************************************************/
+
+var errIllegalFlag error = errors.New("the '-i' flag can only be used with pipes")
 
 /******************************************************************************
 
@@ -71,6 +74,9 @@ func convertCommand(c *cli.Context) error {
 		fmt.Fprint(c.App.Writer, string(output))
 
 	} else {
+		if c.String("i") != "" {
+			return errIllegalFlag
+		}
 
 		// gets glob pattern matches to determine which files to use.
 		matches := getMatches(c)
@@ -141,6 +147,9 @@ func hashCommand(c *cli.Context) error {
 		printHash(c, hash, "-")
 
 	} else {
+		if c.String("i") != "" {
+			return errIllegalFlag
+		}
 
 		// gets glob pattern matches to determine which files to use.
 		matches := getMatches(c)
--- a/poly/commands.go
+++ b/poly/commands.go
@@ -35,7 +35,7 @@ Tim
 
 ******************************************************************************/
 
-var errIllegalFlag error = errors.New("the '-i' flag can only be used with pipes")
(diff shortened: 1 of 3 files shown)


poly/main.go
```diff
@@ -95,7 +95,7 @@ func application() *cli.App {
 					&cli.StringFlag{
 						Name:  "i",
 						Value: "",
-						Usage: "Specify file input type. Options are Gff, gbk/gb, and json. Defaults to none.",
+						Usage: "Specify file input type. Options are gff, gbk/gb, fasta, and json. Defaults to none. For use with pipes.",
```
jkhales, line 98: Another place that maybe good to throw an error is if no valid `-i` value is supplied when piping.

(Also, is `Defaults to none` a valid case?)

TimothyStiles: Definitely not a valid case while piping. You're right that it should be throwing an error in this case @jkhales.

TimothyStiles:
> how should passing -i be handled when the input source is not a pipe?

Unfortunately I think some people write out FASTA without a file extension which is why I did this originally. For our sanity it'd probably be best just to throw an error if someone tries to use the `-i` flag when not piping.


> Currently it is ignored, but probably should throw an error. Unsure how much this repo follows the "fail fast and loudly" philosophy though.

When I first started writing Poly this was one of the first sections I wrote. Wasn't following this philosophy then but am now. Definitely should throw an error going forward.

poly/main.go
```diff
@@ -114,7 +114,7 @@ func application() *cli.App {
 
 					&cli.StringFlag{
 						Name:  "i",
-						Value: "json",
+						Value: "",
```
jkhales, line 117: Removed this default value to avoid triggering an error each time `hash` was called without a pipe.

poly/main.go
```diff
@@ -111,12 +101,7 @@ func application() *cli.App {
 				Usage:   "Hash a sequence while accounting for circularity using SeqhashV1.",
 
 				Flags: []cli.Flag{
-
-					&cli.StringFlag{
-						Name:  "i",
-						Value: "json",
```
jkhales, line 117: Removed this default value to avoid triggering an error each time `hash` was called without a pipe.

poly/main.go
```diff
@@ -50,28 +50,28 @@ func run(args []string) {
...
 func application() *cli.App {
+	inputFlag := cli.StringFlag{
+		Name:  "i",
+		Usage: "Specify file input type. Options are gff, gbk/gb, fasta, and json. For use with pipes.",
+	}
+	outputFlag := cli.StringFlag{
+		Name:  "o",
+		Value: "json",
+		Usage: "Specify file output type or path. Defaults to json.",
+	}
```
jkhales, line 61: I noticed that the input and output flags were the same for the top level and the sub commands.

I consolidated them here and standardized the help text.

jkhales:
Hey @TimothyStiles, I added some errors and tests.

I also consolidated some code duplication.

poly/commands.go
```diff
@@ -34,6 +35,8 @@ Tim
 
 ******************************************************************************/
 
+var errIllegalFlag error = errors.New("the '-i' flag can only be used with pipes")
```
TimothyStiles, line 38: `errIllegalFlag` should be `errIllegalInputFlag` just in case there is ever any other illegal flags we should be throwing errors for.

TimothyStiles (commented):
There's only one minor change needed which is just renaming a variable. After that we'll be set to merge @jkhales!

TimothyStiles:
Merged @jkhales!

PR #149, opened 2021-05-15
Make writes to temporary files in io tests

jkhales:
I noticed that not all files were cleaned up after running `go test`:

Screen Shot 2021-05-14 at 7 52 17 PM

* Fixed this by adding in `defer os.Remove` calls next to the writes so it's easier to keep track of cleanup.
* Switched the save location to a `TempDir` so even in the case of a catastrophic failure, the file system will take care of clean up. This also isolates the test correctly so there won't be any chance of file name conflict.


io_test.go
```diff
@@ -151,63 +162,89 @@ func ExampleBuildGbk() {
...
+
+func TestGbLocationStringBuilder(t *testing.T) {
+	tmpDataDir, err := ioutil.TempDir("", "data-*")
+	if err != nil {
+		t.Error(err)
+	}
+	defer os.RemoveAll(tmpDataDir)
 
-	scrubbedGbk = ReadGbk("data/t4_intron.gb")
+	scrubbedGb := ReadGbk("data/t4_intron.gb")
```
TimothyStiles, line 232: lol

TimothyStiles (approved):
Overall this looks good and can be approved. Not sure why gbk is now gb but ¯\_(ツ)_/¯

Issue #134, opened 2021-05-04
Empty codon tables return panic instead of error when optimizing


Koeng101:
**Describe the bug**
If you pass an empty codon table into optimize, you get the following error:
```
--- FAIL: TestFixCds (0.00s)
panic: invalid argument to Intn [recovered]
        panic: invalid argument to Intn

goroutine 14 [running]:
testing.tRunner.func1.2(0x728e00, 0x7c0620)
        /usr/lib/go/src/testing/testing.go:1143 +0x332
testing.tRunner.func1(0xc000297e00)
        /usr/lib/go/src/testing/testing.go:1146 +0x4b6
panic(0x728e00, 0x7c0620)
        /usr/lib/go/src/runtime/panic.go:965 +0x1b9
math/rand.(*Rand).Intn(0xc000104330, 0x0, 0xc0001007e0)
        /usr/lib/go/src/math/rand/rand.go:169 +0x9d
math/rand.Intn(...)
        /usr/lib/go/src/math/rand/rand.go:337
github.com/mroth/weightedrand.Chooser.Pick(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/koeng/go/pkg/mod/github.com/mroth/weightedrand@v0.2.1/weightedrand.go:48 +0x3b
github.com/TimothyStiles/poly.Optimize(0x787c82, 0x35b, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/koeng/go/src/github.com/TimothyStiles/poly/transformations.go:112 +0x299
github.com/TimothyStiles/poly.TestFixCds(0xc000297e00)
        /home/koeng/go/src/github.com/TimothyStiles/poly/synthesis_test.go:68 +0x385
testing.tRunner(0xc000297e00, 0x78a0e0)
        /usr/lib/go/src/testing/testing.go:1193 +0xef
created by testing.(*T).Run
        /usr/lib/go/src/testing/testing.go:1238 +0x2b3
exit status 2
FAIL    github.com/TimothyStiles/poly   0.222s
```

Instead, you should get an error.

TimothyStiles:
@Koeng101 can you add error handling to this? I'd be happy to merge.

PR #151, opened 2021-05-15
Add error handling to Optimize and Translate

jkhales:
Added basic validation and error handling to the two main functions in transformations.go, `Optimize` and `Translate` as requested in:

https://github.com/TimothyStiles/poly/issues/134

It could be nice to add some typing/validation to amino acid and sequence strings that do a regex check.


jkhales:
Also, is `poly` versioned in any sense? This would definitely be a "breaking" change.

TimothyStiles:
Yeah we are @jkhales. If you look at releases on the right side of the project main you can see what we're on. I usually put BREAKING CHANGES on release notes and that seems to do the trick for now.

jkhales:
> Yeah we are @jkhales. If you look at releases on the right side of the project main you can see what we're on. I usually put BREAKING CHANGES on release notes and that seems to do the trick for now.

Nice, so there is no changelog or anything I need to edit? Or is there something else to do for this PR?

TimothyStiles:
@jkhales is there any way to test error handling too? Would be good to have in tests other than just assigning to `_` and ignoring.

rkrishnasanka:
> Yeah we are @jkhales. If you look at releases on the right side of the project main you can see what we're on. I usually put BREAKING CHANGES on release notes and that seems to do the trick for now.

There needs to be a version bump if its breaking changes (minor at least)

jkhales:
> @jkhales is there any way to test error handling too? Would be good to have in tests other than just assigning to `_` and ignoring.

Did you see these tests that I added?

https://github.com/TimothyStiles/poly/pull/151/files#diff-bdccce56349e8b0b47fc7d38f6449396ae832f9491993497189daa63544f5935R32

https://github.com/TimothyStiles/poly/pull/151/files#diff-bdccce56349e8b0b47fc7d38f6449396ae832f9491993497189daa63544f5935R82

transformations.go
```diff
@@ -55,6 +55,10 @@ Tim
 
 ******************************************************************************/
 
+var errEmtpyCodonTable error = errors.New("empty codon table")
+var errEmtpyAminoAcidString error = errors.New("empty amino acid string")
+var errEmtpySequenceString error = errors.New("empty sequence string")
```
TimothyStiles, line 60: removed as global variable since it's only being used once.

PR #156, opened 2021-05-17
Uniprot XML parser

Koeng101:
Initial uniprot XML parser. This basically parses Uniprot XML into each entry. This parser streams in XML rather than processing it all at once, and so its output is a channel. Even Uniprot Sprot's XML is ~6.5gb uncompressed, so you really do need this streaming ability. 

The structs in xml.go were actually automatically generated with [xsdgen](https://pkg.go.dev/aqwari.net/xml/cmd/xsdgen) and the [uniprot.xsd](https://www.uniprot.org/docs/uniprot.xsd) file. 

TODO:

1. Test all of xml.go (even though it was automatically generated)
2. Make docs for `Entry` and perhaps higher level structs for it


Koeng101:
Documented the files, but did not fully test xml.go. Kinda hard to, since it can change.

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,66 @@
...
+	xmlFile, _ := os.Open(path)
+	r, _ := gzip.NewReader(xmlFile)
+	go ParseUniprot(r, entries)
+}
+
+// ParseUniprot parses Uniprot entries into a channel.
+func ParseUniprot(r io.Reader, entries chan<- Entry) {
+	decoder := xml.NewDecoder(r)
+	for {
+		t, _ := decoder.Token()
```
jkhales, line 51: why is there no chance of error here?

jkhales, line 51:
```suggestion
		decoderToken, _ := decoder.Token()
```

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,66 @@
...
+		if t == nil {
+			break
+		}
+		switch se := t.(type) {
```
jkhales, line 55: if there is only one case, an `if` statement makes more sense.

What does `se` stand for?

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,66 @@
...
+		case xml.StartElement:
+			switch se.Name.Local {
+			case "entry":
```
jkhales, line 58:
```suggestion
		se := t.(type) 
		if se == xml.StartElement && se.Name.Local == "entry"{
		...
```
OK at the risk of making the same mistake my review of your other PR

parsers/uniprot/xml.go
```diff
@@ -0,0 +1,585 @@
...
+	Last        string            `xml:"last,attr,omitempty"`
+	Publisher   string            `xml:"publisher,attr,omitempty"`
+	City        string            `xml:"city,attr,omitempty"`
+	Db          string            `xml:"db,attr,omitempty"`
+	Number      string            `xml:"number,attr,omitempty"`
+	Institute   string            `xml:"institute,attr,omitempty"`
+	Country     string            `xml:"country,attr,omitempty"`
+}
+
+// Describes a cofactor.
```
jkhales, line 50:
```suggestion
```
redundantt

parsers/uniprot/xml.go
```diff
@@ -0,0 +1,585 @@
...
+}
+
+// Describes the authors of a citation when these are represented by a consortium.
+// Equivalent to the flat file RG-line.
+type ConsortiumType struct {
+	Name string `xml:"name,attr"`
+}
+
+// May be one of Swiss-Prot, TrEMBL
+type Dataset string
```
jkhales, line 122: should this be an enum?

parsers/uniprot/xml.go
```diff
@@ -0,0 +1,585 @@
...
+// Equivalent to the flat file DR-line.
+type DbReferenceType struct {
+	Molecule string         `xml:"http://uniprot.org/uniprot molecule,omitempty"`
+	Property []PropertyType `xml:"http://uniprot.org/uniprot property,omitempty"`
+	Type     string         `xml:"type,attr"`
+	Evidence IntListType    `xml:"evidence,attr,omitempty"`
+}
+
+// May be one of left-to-right, right-to-left
+type Direction string
```
jkhales, line 136: enum?

parsers/uniprot/xml.go
```diff
@@ -0,0 +1,585 @@
...
+	Original    string       `xml:"http://uniprot.org/uniprot original,omitempty"`
+	Variation   []string     `xml:"http://uniprot.org/uniprot variation,omitempty"`
+	Location    LocationType `xml:"http://uniprot.org/uniprot location"`
+	Type        Type         `xml:"type,attr"`
+	Description string       `xml:"description,attr,omitempty"`
+	Evidence    IntListType  `xml:"evidence,attr,omitempty"`
+}
+
+// May be one of single, multiple
+type Fragment string
```
jkhales, line 233: enum?

parsers/uniprot/xml.go
```diff
@@ -0,0 +1,585 @@
...
+	Evidence IntListType `xml:"evidence,attr,omitempty"`
+}
+
+type Transposon struct {
+	Value    string      `xml:",chardata"`
+	Evidence IntListType `xml:"evidence,attr,omitempty"`
+}
+
+// May be one of apicoplast, chloroplast, organellar chromatophore, cyanelle, hydrogenosome, mitochondrion, non-photosynthetic plastid, nucleomorph, plasmid, plastid
+type Type string
```
jkhales, line 543: enum?

jkhales, line 543: This needs a more descriptive name.

jkhales: Oh just reread the comment and this is just autogenerated code? 🤔 

Is it important that we adhere strictly to the autogeneration or would fixing up some of the naming be reasonable?

parsers/uniprot/xml.go
Koeng101: xml.go is automatically generated, so I don't think we should edit it.

parsers/uniprot/xml.go
Koeng101: We could fix it up, but we might have to autogenerate again, in which case all of our edits would be wiped. So I think it is better to document outside the autogenerated stuff.

parsers/uniprot/uniprot.go
Koeng101: There is a chance of an error if Uniprot serves us invalid XML. We should actually ideally have this as an error, in case there are any problems in the bytes as they get served to us on disk... let me figure out how to do this.

parsers/uniprot/uniprot.go
Koeng101: `.(type)` requires a switch so we cannot use an if statement (part of Golang https://tour.golang.org/methods/16)

startElement. Fixing that.

parsers/uniprot/uniprot.go
Koeng101: Unfortunately type assertions don't work that way in go :(

Koeng101:
Added the check_uniprot script for @TimothyStiles

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,81 @@
...
+		decoderToken, err := decoder.Token()
+		if err != nil {
+			errors <- err
+		}
+		if decoderToken == nil {
+			break
+		}
+		switch startElement := decoderToken.(type) {
+		case xml.StartElement:
+			if startElement.Name.Local == "entry" {
```
jkhales, line 69:
```suggestion
		startElement, ok := decoderToken.(xml.StartElement) {
		if ok && startElement.Name.Local == "entry" {
```

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+	"encoding/xml"
+	"io"
+	"os"
+)
+
+/******************************************************************************
+May 18, 2021
+
+Uniprot is comprehensive, high-quality and freely accessible resource of protein
+sequence and functional information. It the best(1) protein database out there.
```
TimothyStiles, line 14: *it's

parsers/uniprot/uniprot_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"fmt"
+)
+
+func ExampleReadUniprot() {
+	entries, _, _ := ReadUniprot("data/uniprot_sprot_mini.xml.gz")
+
+	entry := <-entries
+	fmt.Println(entry.Accession[0])
+	// Output: P0C9F0
+}
```
TimothyStiles, line 13: Can you write a test for `xml.go`'s Marshall functionality? I realized it'd be faster if you did this @Koeng101.

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+
+// ReadUniprot reads a gzipped Uniprot XML dump.
+func ReadUniprot(path string) (chan Entry, chan error, error) {
+	entries := make(chan Entry)
+	errors := make(chan error)
+	xmlFile, err := os.Open(path)
+	if err != nil {
+		return entries, errors, err
+	}
+	r, err := gzip.NewReader(xmlFile)
```
TimothyStiles, line 48: Should this be default gzip?

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+
+******************************************************************************/
+
+// ReadUniprot reads a gzipped Uniprot XML dump.
+func ReadUniprot(path string) (chan Entry, chan error, error) {
+	entries := make(chan Entry)
+	errors := make(chan error)
+	xmlFile, err := os.Open(path)
+	if err != nil {
+		return entries, errors, err
```
TimothyStiles, line 46: This line isn't covered in tests.

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	r, err := gzip.NewReader(xmlFile)
+	if err != nil {
```
TimothyStiles, line 49: This line isn't covered in tests.

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+		return entries, errors, err
```
TimothyStiles, line 50: why errors and err? Can these be merged or renamed?

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+(2) https://www.uniprot.org/downloads
+(3) https://www.uniprot.org/docs/uniprot.xsd
+
+******************************************************************************/
+
+// ReadUniprot reads a gzipped Uniprot XML dump.
+func ReadUniprot(path string) (chan Entry, chan error, error) {
+	entries := make(chan Entry)
+	errors := make(chan error)
+	xmlFile, err := os.Open(path)
```
TimothyStiles, line 44: be more specific with err name if there are multi errs in function.

TimothyStiles (dismissed):
Errors need to be handled in tests and Marshall functionality needs to be tested for better overall coverage.

parsers/uniprot/uniprot.go
TimothyStiles, line 48: single letter variable. Be more specific.

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+		if decoderToken == nil {
+			break
+		}
+		//switch startElement := decoderToken.(type) {
+		//case xml.StartElement:
+		startElement, ok := decoderToken.(xml.StartElement)
+		if ok && startElement.Name.Local == "entry" {
+			if startElement.Name.Local == "entry" {
+				var e Entry
+				err = decoder.DecodeElement(&e, &startElement)
```
jkhales, line 73: Is this mutating the `decoder`? I looked at the spec and it didn't mention anything: https://golang.org/pkg/encoding/xml/#Decoder.DecodeElement

If it's not than why isn't this an infinite loop?

parsers/uniprot/uniprot.go
```diff
@@ -0,0 +1,83 @@
...
+	return entries, errors, nil
+}
+
+// ParseUniprot parses Uniprot entries into a channel.
+func ParseUniprot(r io.Reader, entries chan<- Entry, errors chan<- error) {
+	decoder := xml.NewDecoder(r)
+	for {
+		decoderToken, err := decoder.Token()
+		if err != nil {
+			errors <- err
```
jkhales, line 62: should this end the for loop? Maybe at least a `continue`?

parsers/uniprot/uniprot.go
Koeng101: Yes, kinda. You take out tokens from the decoder as it goes along, and then stop once the decoder runs out of tokens.

parsers/uniprot/uniprot.go
Koeng101: Yes.

parsers/uniprot/uniprot.go
Koeng101: Fixed.

parsers/uniprot/uniprot_test.go
Koeng101: No. Just removed em

parsers/uniprot/uniprot.go
Koeng101: Fixed.

Koeng101: Also I can see what lines aren't covered in tests

parsers/uniprot/uniprot.go
Koeng101: No, but I added comments on them to clarify why we have 2.

parsers/uniprot/uniprot.go
Koeng101: Resolved.

parsers/uniprot/uniprot.go
Koeng101: No, we should log the error and continue along (which the control flow currently does)

Koeng101: I added continue to make it clearer. Should do that!

parsers/uniprot/uniprot_test.go
```diff
@@ -0,0 +1,29 @@
...
+	entries, _, _ := ReadUniprot("data/uniprot_sprot_mini.xml.gz")
+
+	var entry Entry
+	for singleEntry := range entries {
+		entry = singleEntry
+	}
+	fmt.Println(entry.Accession[0])
+	// Output: O55723
+}
+
```
TimothyStiles, line 18: Can you add an example test form ParseUniprot?

Koeng101: Done!

Issue #145, opened 2021-05-07
The Poly RBS calculator


Koeng101:
Salis lab has previously made a ribosomal binding site calculator, which can predict translation initiation rates from proteins.

However, it is slow (requiring a queue on a website) and closed source. In order to incorporate RBS calculation data in more complex applications, we need better performance and velocity of development. The best advancements in technology should be incorporated in an open-source manner.

### Basic idea of RBS calculator

The basic idea behind the RBS calculator (this is a simplification) is that you take the binding energy of the the ribosomal 16S RNA to the mRNA's RBS site and subtract that from the binding energy of the mRNA to itself. There are a few other variables, but these are the basic ones (please check https://pubs.acs.org/doi/suppl/10.1021/acssynbio.0c00394/suppl_file/sb0c00394_si_001.pdf table 2 for equations)

mRNA is a large variable, so it must be calculated each time the simulation is run. The 16S RNA, on the other hand, is not very variable. There is approximately a power-law distribution of what organisms people use, so we can cache most of the 16S-RNA to RBS (which I will now call 16S-RBS) data in a lookup table. 

### Software and numbers we need

It is important to keep in mind we want this software to be **fast**. In order to get performance, there are 2 primary optimizations: 1 - using a faster algorithm for calculating RNA secondary structure (we use LinearFold, which folds RNA in linear time) and 2 - using a lookup table for slow RNA-to-RNA binding calculations.

In order to calculate mRNA folding, @vivekr has ported [LinearFold](https://doi.org/10.1093/bioinformatics/btz375) to [Golang](https://github.com/vivkr/linearfold). This package needs to be incorporated into Poly before we build the RBS calculator.

In order to calculate the 16S-RBS lookup table, we will likely need to operate outside of Golang (probably in python). LinearFold does not support (at this time) multiple separate RNAs binding to each other, so we'll have to do this work in a different algorithm. It will be a challenge to relate the two numbers from different software packages. Since the 16S RNA binding sequence is only 9 base pairs long, we theoretically only have to calculate its binding efficiency to 262,144 other RNAs.

There are other parameters that assist in doing RBS calculations (such as ΔGstandb, from https://pubs.acs.org/doi/suppl/10.1021/acssynbio.0c00394/suppl_file/sb0c00394_si_001.pdf). We'll likely need to build those into the calculator at some point, but perhaps not in version 1. 

### Testing

[...]

PR #161, opened 2021-05-21
Add ability to calculate minimum free energy

v-raja:
This PR adds the ability to calculate minimum free energy of a given structure (in dot-bracket notation) and its sequence (needed for #145 ). The calculations compute the correct results for the given tests. The code still needs to be cleaned up and documented well.

@TimothyStiles I have to include a file for the energy parameters needed for the calculation. I was wondering if you'd prefer the three mfe files (`mfe.go`, `mfe_test.go` and `energy_params.go`) to be put in a folder, or leave them as is?


vivkr:
@TimothyStiles This PR is ready for review.

I've documented the code the best I can and only two functions need to be documented now (which I'll work on, but may take a while). Till then though, the rest of the `mfe.go` and `mfe_test.go` code is ready to be reviewed. 

I need to work on `energy_params.go`, but that's just a lot of renaming variables and formatting the code to Go's style so should be done tomorrow.

mfe.go
```diff
@@ -0,0 +1,1092 @@
...
+	ninio                                     [5]int
+	lxc                                       float64
+	MLbase, MLclosing, TerminalAU, DuplexInit int
+	MLintern                                  [nbPairs + 1]int
+	Tetraloops                                string
+	Tetraloop                                 [200]int
+	Triloops                                  string
+	Triloop                                   [40]int
+	Hexaloops                                 string
+	Hexaloop                                  [40]int
```
Koeng101, line 183: Why are some attributes private and some public?

mfe.go
```diff
@@ -0,0 +1,1092 @@
...
+	return energy, EnergyContribution{energy: energy}
+}
+
+/**
+ *  Evaluate a stem branching off the exterior loop.
+ *
+ *  Given a base pair (i,j) (encoded by `basePairType`), compute the
+ *  energy contribution including dangling-end/terminal-mismatch contributions.
+ *  Instead of returning the energy contribution per-se, this function returns
+ *  the corresponding Boltzmann factor.
```
Koeng101, line 346: What is a Boltzmann factor? Can we get a small review of that here to understand what it is returning?

mfe.go
```diff
@@ -0,0 +1,1092 @@
...
+
+	if basePairType > 2 {
+		// The closing base pair is not a GC or CG pair (could be GU, UG, AU, UA, or non-standard)
+		energy += energyParams.TerminalAU
+	}
+
+	return energy
+}
+
+// TODO: document
```
Koeng101, line 379: ^

mfe.go
```diff
@@ -0,0 +1,1092 @@
...
+
+/**
+ *  Compute the energy contribution of a multi-loop stem.
+ *
+ *  Given a base pair (i,j) (encoded by `basePairType`), compute the
+ *  energy contribution including dangling-end/terminal-mismatch contributions.
+ *  You can prevent taking 5'-, 3'-dangles or mismatch contributions into
+ *  account by passing -1 for `fivePrimeMismatch` and/or `threePrimeMismatch`
+ *  respectively.
+ *
```
Koeng101, line 833: How hard would it be to make a diagram like those above? They were very helpful.

mfe_test.go
```diff
@@ -0,0 +1,1270 @@
...
+		mfe, energyContributions, _ := CalculateMFE(sequence, structure, DefaultTemperature)
+		logEnergyContributions(energyContributions, sequence)
+		log.Printf("%v", mfe)
+	})
+
+	if stripWhiteSpace(output) != stripWhiteSpace(expected_output) {
+		t.Errorf("\n\nFailed to calcualte mfe for '%v'. \nExpected: \n'%v'\n\nGot: \n'%v'\n\n",
+			sequence, expected_output, output)
+	}
+}
```
Koeng101, line 1255: Lets also include an example of a realistic function that a user might use.

mfe.go
```diff
@@ -0,0 +1,1092 @@
...
+	lenSequence := len(sequence)
+	encodedSequence := make([]int, lenSequence)
+
+	// encode the sequence based on nucleotideRuneMap
+	for i := 0; i < lenSequence; i++ {
+		encodedSequence[i] = nucleotideRuneMap[sequence[i]]
+	}
+
+	return encodedSequence
+}
```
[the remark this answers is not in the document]
v-raja: As we discussed, since LinearFold encodes nucleotides with a different map ('A' -> 0), we can't encode them with the same function. 

(Changing the MFE calculation code to use the same map as LinearFold would be a lot of work as this map dictates where the relevant energy values are found in the energy parameters matrices. So we'd have to change all those matrices if we change this mapping.)

mfe.go
v-raja: Their field names didn't follow a specific convention with regards to capitalisation. So the public fields just happen to be field names which they capitalised. Good catch - will rename to make them all private.

mfe.go
v-raja: I tried to read up about the Boltzmann factor in their repo and it looks like it has nothing to do with this function. The comment relates to another function, but not sure why it's included here. I've removed it now.

mfe_test.go
v-raja: I added the test `ExampleMinimumFreeEnergy()`. Could you confirm that it's a realistic example?

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+
+// Contains all the energy parameters needed for the free energy calculations
+type energyParams struct {
+	stackingPair [nbPairs + 1][nbPairs + 1]int
+	hairpinLoop  [31]int
+	bulge        [maxLoop + 1]int
+	/* This field was previous called internal_loop, but has been renamed to
+	interior loop to remain consistent with the names of other interior loops */
+	interiorLoop                                       [maxLoop + 1]int
+	mismatchExteriorLoop                               [nbPairs + 1][5][5]int
```
TimothyStiles, line 235: What are these magic number indices?

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+	dangle3                                            [nbPairs + 1][5]int
+	interior1x1Loop                                    [nbPairs + 1][nbPairs + 1][5][5]int
+	interior2x1Loop                                    [nbPairs + 1][nbPairs + 1][5][5][5]int
+	interior2x2Loop                                    [nbPairs + 1][nbPairs + 1][5][5][5][5]int
+	ninio                                              [5]int
+	lxc                                                float64
+	multiLoopBase, multiLoopClosingPenalty, terminalAU int
+	multiLoopIntern                                    [nbPairs + 1]int
+	tetraloops                                         string
+	tetraloop                                          [200]int
```
TimothyStiles, line 251: magic number

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+		'C': nucelotideCEncodedTypeMap,
+		'G': nucelotideGEncodedTypeMap,
+		'U': nucelotideUEncodedTypeMap,
+	}
+
+	return basePairMap
+}
+
+// Contains all the energy parameters needed for the free energy calculations
+type energyParams struct {
```
TimothyStiles, line 228: parameterize as many magic numbers as possible in here.

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+
+	if stackIdx != 0 {
+		return nil, fmt.Errorf("%v\nunbalanced brackets '%v%v' found while extracting base pairs",
+			structure, openBracket, closeBracket)
+	}
+
+	return pairTable, nil
+}
+
+func evaluateFoldCompound(fc *foldCompound) (int, []EnergyContribution) {
```
TimothyStiles, line 316: comment the heck out of this function and control flow.

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+
+	// the characters of the opening and closing brackets in the structure string
+	var openBracket, closeBracket byte = '(', ')'
+
+	// keeps track of the indexes of open brackets. Indexes of open brackets are
+	// pushed onto stack and poped off when a closing bracket is encountered
+	var openBracketIdxStack []int = make([]int, lenStructure)
+	var stackIdx int = 0
+
+	for i := 0; i < len(structure); i++ {
```
TimothyStiles, line 285: what are you indexing here for I?

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+
+// Returns the minimum of two ints
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func scaleEnergyParams(temperature float64) *energyParams {
```
TimothyStiles, line 991: This function's control flow is absolute chaos and we need a ton of comments to explain what's going on here.

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+							int21_dH[i][j][k][l][m],
+							tempf)
+					}
+				}
+			}
+		}
+	}
+
+	/* interior 2x2 loops */
+	for i := 0; i <= nbPairs; i++ {
```
TimothyStiles, line 1146: Why so many nests?

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+	}
+
+	params.tetraloops = Tetraloops
+	params.triloops = Triloops
+	params.hexaloops = Hexaloops
+
+	return params
+}
+
+func rescaleDg(dG, dH, dT int) int {
```
TimothyStiles, line 1170: What does this do exactly?

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+ *  @param  fivePrimeMismatch    The 5'-mismatching encoded nucleotide
+ *  @param  threePrimeMismatch   The 3'-mismatching encoded nucleotide
+ *  @param  sequence						 The sequence of the loop
+ *  @param  energyParams     		 The datastructure containing scaled energy parameters
+ *  @return The free energy of the hairpin loop in dcal/mol
+ */
+func evaluateHairpinLoop(size, basePairType, fivePrimeMismatch, threePrimeMismatch int, sequence string, energyParams *energyParams) int {
+	var energy int
+
+	if size <= 30 {
```
TimothyStiles, line 798: parameterize magic numbers in control flow.

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+	if basePairType > 2 {
+		// The closing base pair is not a GC or CG pair (could be GU, UG, AU, UA, or non-standard)
+		energy += energyParams.terminalAU
+	}
+
+	return energy
+}
+
+// Returns the minimum of two ints
+func min(a, b int) int {
```
TimothyStiles, line 984: lol golang only having min for floats.

mfe/mfe.go
```diff
@@ -0,0 +1,1209 @@
...
+		multipleCA:              rescaleDg(MultipleCA37, MultipleCAdH, tempf),
+		multipleCB:              rescaleDg(TerminalAU37, TerminalAUdH, tempf),
+		terminalAU:              rescaleDg(TerminalAU37, TerminalAUdH, tempf),
+		multiLoopBase:           rescaleDg(ML_BASE37, ML_BASEdH, tempf),
+		multiLoopClosingPenalty: rescaleDg(ML_closing37, ML_closingdH, tempf),
+	}
+
+	params.ninio[2] = rescaleDg(ninio37, niniodH, tempf)
+
+	for i := 0; i < 31; i++ {
```
TimothyStiles, line 1006: What are you iterating for all `i`s here?

mfe/mfe_test.go
```diff
@@ -0,0 +1,1271 @@
...
+		Interior loop (2632,2652) GC; (2634,2646) AU:   510
+		Interior loop (2634,2646) AU; (2635,2645) CG:  -220
+		Interior loop (2635,2645) CG; (2636,2644) GC:  -240
+		Interior loop (2636,2644) GC; (2637,2643) GU:  -150
+		Hairpin  loop (2637,2643) GU              :   520
+		Multi    loop (2275,2662) AU              :  -430
+		-936.5`, t)
+}
+
+func test(sequence, structure string, expected_output string, t *testing.T) {
```
TimothyStiles, line 1245: More specific name than `test` is needed here.

mfe/mfe_test.go
```diff
@@ -0,0 +1,1271 @@
...
+
+func ExampleMinimumFreeEnergy() {
+	mfe, _, _ := MinimumFreeEnergy("ACGAUCAGAGAUCAGAGCAUACGACAGCAG", "..((((...))))...((........))..", DefaultTemperature)
+	fmt.Println(mfe)
+	// Output:
+	// -2.9
+}
+
+func TestMinimumFreeEnergy(t *testing.T) {
+	test("ACGAUCAGAGAUCAGAGCAUACGACAGCAG",
```
TimothyStiles, line 20: Can you comment what conditions these tests are testing for?

TimothyStiles (changes requested):
This is an absolutely spectacular PR. I'm calling in help to review this because even with how amazing it is we're going to need some more eyeballs to make this immaculate.

🚀 🚀 🚀

mfe/mfe.go
v-raja: What do you mean by parameterize?

TimothyStiles: Give them variable assignments. numbers should be assigned to variables.

mfe/mfe_test.go
v-raja: What do you mean by conditions?

mfe/mfe.go
v-raja: Not sure what we can do about this. We have to iterate through every value of the matrix to set it.

mfe/mfe.go
v-raja: Do we? We're just setting the values for the energy parameters so we're just looping through the matrices and setting values. Not sure if the control flow needs more information.

mfe/mfe.go
```diff
@@ -0,0 +1,1357 @@
...
+		foldCompound - holds all information needed to compute the free energy of a
+			RNA secondary structure (a RNA sequence along with its folded structure).
+		EnergyContribution - holds information about the energy contribution of a
+			single loop, and the indexes of the base pairs that delimit the loop in
+			the RNA secondary structure.
+
+	Big functions from this file:
+
+		MinimumFreeEnergy - given a RNA sequence, it's structure, and a temperature,
+			it returns the minimum free energy of the RNA secondary structure at the
```
jkhales, line 41: Given a secondary structure and specified temperature, is the "free energy" not determined? If it is determined then: 
```suggestion
			it returns the free energy of the RNA secondary structure at the
```
If not, out of curiosity, what else factors into it?

Also, what is the difference between _free_ energy and just energy?

jkhales (commented):
I cloned the branch locally and believe the energy_params.go should be a const instead of a var?

mfe/mfe.go
```diff
@@ -0,0 +1,1357 @@
...
+ *  energy contribution of each loop.
+ *
+ *  @param sequence         A RNA sequence
+ *  @param structure        Secondary structure in dot-bracket notation
+ *  @param temperature      Temperature at which to evaluate the free energy of the structure
+ *  @return                 The free energy of the input structure given the input sequence in kcal/mol,
+														and a slice of `EnergyContribution` (which gives information about the
+														energy contribution of each loop present in the secondary structure)
+*/
+func MinimumFreeEnergy(sequence, structure string, temperature float64) (float64, []EnergyContribution, error) {
```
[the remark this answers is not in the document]
Koeng101: http://rna.tbi.univie.ac.at/RNAWebSuite/help.html 

MFE is the term used in the field for this calculation. Technically it is "minimum of free energy", but they call it "Minimum Free Energy" (or mfe). If you check out the big RNA folding libraries, you see stuff like http://rna.tbi.univie.ac.at/cgi-bin/RNAWebSuite/RNAfold.cgi , so I think this one is appropriate to keep so long as it has a good comment on it.

TimothyStiles:
@v-raja Is this PR stale given #176?

v-raja:
Nope. #176 parses energy param files into an `EnergyParams` struct. This PR actually figures out the secondary structure and uses the relevant energy values from the `EnergyParams` struct to calculate the minimum free energy.

Koeng101:
Closed in favor of #173

Issue #117, opened 2021-04-27
Standard VScode and GH codespace setup.


TimothyStiles:
**Is your feature request related to a problem? Please describe.**
I believe VScode and Github codespaces allow you to share settings and plugins within `.vscode` directory

**Describe the solution you'd like**
I should collect the tools that I use personally and include them as part of repo settings in `.vscode`.

rkrishnasanka:
What all do you need working on this? I've been playing around with this for my tools which are far more complex in terms of dependency chains, I can probably get a draft of this done this weekend.

TimothyStiles:
@rkrishnasanka At the very minimum something that includes the Go extension published by microsoft with code coverage highlighting turned on. I've also been experimenting the official vision studio liveshare extension which could be of use but I still don't really know how it works in full.

TimothyStiles:
Now that I'm thinking of it perhaps the official github and github code extensions would be nice but probably not necessary to include? I'd also be curious to hear what other people are using for extensions.

rkrishnasanka:
> @rkrishnasanka At the very minimum something that includes the Go extension published by microsoft with code coverage highlighting turned on. I've also been experimenting the official vision studio live share extension which could be of use but I still don't really know how it works in full.

Basically, they create a docker image with the code folder mounted, vscode installed. It then uses the remote dev features (which are really good) to connect to the docker image and runs stuff.

TimothyStiles:
Closing as this has been solved as of #171.

PR #171, opened 2021-06-07
Enables Dev Container #117

rkrishnasanka:
Added vscode dev container settings. Should install the go package a couple of other vscode plugins. Currently, the dev container image uses a Ubuntu Focal Container and installs golang using the latest apt package `golang`.

This dev container also installs the following plugins:
1. `golang.go`
2. `ms-vsliveshare.vsliveshare-pack`
3. `redhat.vscode-yaml`
4. `yzhang.markdown-all-in-one`
5. `premparihar.gotestexplorer`
6. `github.vscode-pull-request-github`
7. `streetsidesoftware.code-spell-checker`
8. `eamodio.gitlens`

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.166.1/containers/ubuntu/.devcontainer/base.Dockerfile
+
+# [Choice] Ubuntu version: bionic, focal
+ARG VARIANT="focal"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
+
+# [Optional] Uncomment this section to install additional OS packages.
+RUN export DEBIAN_FRONTEND=noninteractive \
+  && sudo apt-get update \
+  && sudo apt-get -y install golang
+
+RUN go get -v golang.org/x/tools/gopls
+RUN go get -u github.com/lukehoban/go-outline
+RUN go get -u github.com/uudashr/gopkgs/cmd/gopkgs
+RUN go get github.com/go-delve/delve/cmd/dlv
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,27 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.166.1/containers/ubuntu
+{
+  "name": "poly - Ubuntu",
+  "build": {
+    "dockerfile": "Dockerfile",
+    // Update 'VARIANT' to pick an Ubuntu version: focal, bionic
+    "args": { "VARIANT": "focal" }
+  },
+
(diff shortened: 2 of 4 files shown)


TimothyStiles:
@rkrishnasanka I absolutely love the extensions you've found here. Does this also work with setting up local VScode environments?

rkrishnasanka:
@TimothyStiles this is only for local VSCode Containers. I'm still figuring out how to make the webversions stick.

rkrishnasanka:
@TimothyStiles Looks like the github codespaces will base off the initial version based on vscode container settings

TimothyStiles:
> @TimothyStiles this is only for local VSCode Containers. I'm still figuring out how to make the webversions stick.

Awesome. We should probably include getting set up with this in the contributor's guidelines and I'll happily be the beta tester for that.

TimothyStiles:
@rkrishnasanka testing out now. Couple of thoughts at first glance.

1. Installation time takes a while. At least a couple of minutes. Will it cache this or do it every time?

2. Docker container actually disconnected from VS code when I tried to open files and fails to render code. Pretty much has crashed and is unusable on my end. Also depends on the fact that end user already has docker installed on their machine. Probably not an issue for shipping on codespaces but looks like a headache install wise locally.

3. Localized containers won't run code that isn't pushed to remote git server. Huge slowdown with dev cycle time.

Is there a way to just ship extensions that we use as a single configured package and not worry about shipping a containerized environment? Go for the most part is stable in this sense and it may be more of a headache troubleshooting docker than just having devs install Go and run in their local environment.

rkrishnasanka:
> @rkrishnasanka testing out now. Couple of thoughts at first glance.
> 
> 1. Installation time takes a while. At least a couple of minutes. Will it cache this or do it every time?

Well it'll only build it the first time it should be fine after that/I can check to make sure that this isn't the case all the time, try connecting it to it again, it wont go through docker build cycle

> 2. Docker container actually disconnected from VS code when I tried to open files and fails to render code. Pretty much has crashed and is unusable on my end. Also depends on the fact that end user already has docker installed on their machine. Probably not an issue for shipping on codespaces but looks like a headache install wise locally.

Do you have a screenshot of what happened ?

> 3. Localized containers won't run code that isn't pushed to remote git server. Huge slowdown with dev cycle time.
> 

Can you clarify what you mean by this? Basically, the way you use dev containers is by mounting whatever files you need onto the docker image so there should technically be no slowdown (vscode will mount your directory onto this docker image)

> Is there a way to just ship extensions that we use as a single configured package and not worry about shipping a containerized environment? Go for the most part is stable in this sense and it may be more of a headache troubleshooting docker than just having devs install Go and run in their local environment.

I don't think so...but there might be something that can be done here. I'm curious to see what your crash looks like

rkrishnasanka:
I've added the ability to save the recommended extensions too

Jun-10-2021 00-02-11

rkrishnasanka:
@TimothyStiles what kind of documentation would you want for this?

TimothyStiles:
@rkrishnasanka I love all of this and I'm absolutely merging it!

PR #174, opened 2021-06-25
Repackage poly into subpackages.

TimothyStiles:
Poly has gotten really big and it makes sense to repackage a lot of it into subpackages.

This PR does exactly that. It runs and all tests pass. If anyone could review it for syntax, clarity, and overall project structure that'd be great.

Before merging I'm going to work on adding go docs for each subpackage with runable examples that will sort of function like integration tests where appropriate and give end users an idea of how all this stuff ties together.

-Tim

EDIT: Should also find a way to make test data more manageable without duplication.


isaacguerreir (approved):
Didn't found any problem.

The code is more readable, definitly. Thanks, Tim!

io/genbank/genbank_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+func ExampleParse() {
+	file, _ := ioutil.ReadFile("../../data/puc19.gbk")
+	sequence := Parse(file)
+
+	fmt.Println(sequence.Meta.Locus.ModificationDate)
+	// Output: 22-OCT-2019
+}
+
+func ExampleBuildGbk() {
```
TimothyStiles, line 36: Should just be ExampleBuild()

io/pson/pson.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Read reads an poly.Sequence JSON file.
+func Read(path string) poly.Sequence {
+	file, _ := ioutil.ReadFile(path)
+	sequence := Parse(file)
+	return sequence
+}
+
+// Write writes an poly.Sequence struct out to json.
```
TimothyStiles, line 36: *writes a*

io/genbank/genbank.go
```diff
@@ -461,7 +93,7 @@ func ParseGbk(file []byte) Sequence {
 }
 
 // BuildGbk builds a GBK string to be written out to db or file.
-func BuildGbk(sequence Sequence) []byte {
+func BuildGbk(sequence poly.Sequence) []byte {
```
TimothyStiles, line 96: should be renamed to `Build`.

clone/clone.go
```diff
@@ -53,8 +57,8 @@ in multiple places (Type IIG enzymes) such as BcgI.
 
 ******************************************************************************/
 
-// CloneSequence is a simple struct that can carry a circular or linear DNA sequence.
-type CloneSequence struct {
+// Part is a simple struct that can carry a circular or linear DNA sequence.
+type Part struct {
```
jkhales, line 61:
```suggestion
type DNASequence struct {
```
Part is maybe too generic

docs/library-io.md
```diff
@@ -173,7 +173,7 @@ That's it. The reason we don't have a `ParseJSON()` is that golang, like almost
   // generating a gff string that then can be piped to stdout or written to a database.
   ecoliGffString := BuildGff(ecoliAnnotatedSequence)
 
-  ecoliGbkString := BuildGbk(ecoliAnnotatedSequence)
+  ecoliGbkString := Build(ecoliAnnotatedSequence)
```
jkhales, line 176: There is a `BuildGff` right above this. Naming is inconsistent/confusing.

Maybe make a universal `Build(..., 'GBK')` interface?

transform/reverse/reverse_test.go
```diff
@@ -0,0 +1 @@
+package reverse
```
jkhales, line 1: Either add tests or delete this?

transform/codon/codon.go
```diff
@@ -70,15 +71,15 @@ type AminoAcid struct {
 	Codons []Codon `json:"codons"`
 }
 
-// CodonTable holds information for a codon table.
-type CodonTable struct {
+// Table holds information for a codon table.
+type Table struct {
```
jkhales, line 75: `Table` seems less descriptive and grep'able than `CodonTable`.

jkhales (approved):
✨

docs/library-io.md
TimothyStiles: I was actually going to remove these docs for good with this PR. Too likely to break and not helpful. Should be replaced with new tutorials that are coming in new PR.

> Maybe make a universal Build(..., 'GBK') interface?

Possibly but I really like the current interface which is `genbank.Build(Sequence)`

transform/codon/codon.go
TimothyStiles: That's true but when used in this new subpackage configuration it's written as `codon.Table` in packages that depend on it so it'd likely be just as grepable. Does feel super weird in package though. I've got similar feelings about `random.ProteinSequence`.

clone/clone.go
TimothyStiles: This was on @Koeng101's suggestion. It was originally called `CloneSequence` but would end up being written as `clone.CloneSequence` when used from another package. I couldn't have it be named `clone.Sequence` since it'd get confusing because the main poly package contains the `Sequence` struct for annotated sequences. I suggested `clone.Vector` but semantically `clone.Part` is most correct. I'll add some DNA part terminology to the comments to clarify this for future devs.

transform/reverse/reverse_test.go
TimothyStiles: I'll add tests. That was on my todo list thanks for catching this and reminding me.

TimothyStiles: Just added a test. Done!

clone/clone.go
Koeng101: I think Part is the most correct domain-wise, but I think jkhales is right that it is too generic for folks not building DNA. I would say keep Part but add a much larger comment.

transform/reverse/reverse_test.go
TimothyStiles: Wait no. Test failed. That's what I get for using GitHub copilot. Was just excited that it wrote the test for me 😅

Koeng101: Should this just be in a general utils package?

transform/codon/codon.go
Koeng101: Yea probably fine now that we have subpackages.

jkhales: ah, if it's always written out as `codon.Table` that is pretty nice.

checks/checks.go
```diff
@@ -0,0 +1,9 @@
+package checks
+
+import "github.com/TimothyStiles/poly/transform/reverse"
+
+// IsPalindromic accepts a sequence of even length and returns if it is
+// palindromic. More here - https://en.wikipedia.org/wiki/Palindromic_sequence
+func IsPalindromic(sequence string) bool {
+	return sequence == reverse.Complement(sequence)
+}
```
Koeng101, line 9: I think we might want to merge this into a utils package with basic complementation and reverse stuff. Thoughts?

Koeng101: Also I don't know how to fix this, but if I remember correctly you code copy paste rather than the copy-paste-file then remove-content, you remove the original contributors "ownership" or whatever git tracks. I don't care, but other folks might.

io/example_test.go
```diff
@@ -0,0 +1,34 @@
+package io
+
+import (
+	"github.com/TimothyStiles/poly/io/fasta"
+	"github.com/TimothyStiles/poly/io/genbank"
+	"github.com/TimothyStiles/poly/io/gff"
+	"github.com/TimothyStiles/poly/io/pson"
```
Koeng101, line 7: What is pson?

transform/reverse/reverse_test.go
jkhales: I'm jealous, extremely impatiently waiting on copilot access.

image

io/pson/pson.go
```diff
@@ -0,0 +1,46 @@
+package pson
```
Koeng101, line 1: Is PSON an alternative to JSON or something? P isn't really around J so I'm wondering if this is not a typo.

transform/variants/variants.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"errors"
+	"strings"
+)
+
+// AllVariantsIUPAC takes a string as input
+// and returns all iupac variants as output
+func AllVariantsIUPAC(seq string) ([]string, error) {
+	seqVariantList := [][]rune{}
+	seqVariants := []string{}
```
Koeng101, line 12: Did this remove contribution?

transform/reverse/reverse_test.go
TimothyStiles: > Should this just be in a general utils package?

Go frowns upon this practice a lot and I tend to agree. Easy to dump a bunch of unrelated stuff in there. I do think the naming for this should be a lot smarter cause it turns out that co-pilot actually wrote a test for a function that returns a complement which without the package name is pretty much what `reverse.Complement` looks like.

io/example_test.go
TimothyStiles: `poly json` thought I wrote a comment about it but essentially the namespace collisions with Go's `json` package was too much of a pain. Should make a comment if there isn't already

transform/reverse/reverse_test.go
Koeng101: ok, that's fair! If it's the go way, might as well

io/pson/pson.go
TimothyStiles: > poly json thought I wrote a comment about it but essentially the namespace collisions with Go's json package was too much of a pain. Should make a comment if there isn't already

transform/variants/variants.go
TimothyStiles: It shouldn't have? I think historically it has done that but git seemed to be tracking where code was going. I'll look into it.

checks/checks.go
TimothyStiles: > I think we might want to merge this into a utils package with basic complementation and reverse stuff. Thoughts?

See -> https://github.com/TimothyStiles/poly/pull/174#discussion_r661935919

TimothyStiles: > Also I don't know how to fix this, but if I remember correctly you code copy paste rather than the copy-paste-file then remove-content, you remove the original contributors "ownership" or whatever git tracks. I don't care, but other folks might.

I worried about this but GitHub seemed to be tracking this pretty well and it may be a new feature of git. I'll look into this before merge.

transform/reverse/reverse_test.go
TimothyStiles: Also tricky thing about `reverse.complement` is that it's used almost everywhere and was causing serious cyclic dependency issues to the point where it kind of needed its own package that didn't rely on anything else. This is probably the case with a lot of util like functions.

io/pson/pson.go
TimothyStiles: See -> https://github.com/TimothyStiles/poly/pull/174#discussion_r661936455

io/example_test.go
jkhales: FWIW, `pson` is definitely a clever name, but `polyJson` maybe a clearer name and avoid confusion such as experienced by @Koeng101 I also didn't realize it was a json wrapper.

Koeng101: I think polyJson would be better

transform/reverse/reverse_test.go
TimothyStiles: Just fixed this in latest push. I moved `ReverseComplement` to the `transform` package so now you call it with `transform.ReverseComplement`. I also pushed `Reverse` and `Complement` functions and Example tests. @Koeng101 @jkhales

io/example_test.go
TimothyStiles: > I think polyJson would be better

@Koeng101 @jkhales Just changed to polyjson in latest push.

TimothyStiles:
@Koeng101 @jkhales @isaacguerreir I'm planning to merge this by the end of today if there's no major objections. I looked into and it seems that I'll have to make a new PR to make sure git blame doesn't get borked and all previous contributors work will be credited but otherwise I think this is ready for a final review.

Also, I've asked @Koeng101 to write an explanation of DNA parts for `clone.Part` as requested by @jkhales that will be included in a new pull request.

Koeng101 (approved):
I approve of this repackage. From my look at the files, I don't see anything that is off.

PR #175, opened 2021-07-01
Add ability to parse Secondary Structure from dot bracket notation

v-raja:


secondary_structure/dot_bracket.go
```diff
@@ -0,0 +1,533 @@
...
+// and `SecondaryStructure` of a RNA sequence
+type parseCompound struct {
+	length             int   // length of `sequence`
+	pairTable          []int // (see `pairTable()`)
+	annotatedStructure []byte
+}
+
+// FromDotBracket returns the annotated structure and `SecondaryStructure` of
+// a RNA sequence from its 'dot-bracket' structure.
+func FromDotBracket(dotBracketStructure string) (string, *SecondaryStructure, error) {
```
Koeng101, line 58: The exported function, FromDotBracket, returns a string, a secondary structure, and an error. In the tests, I see only the string output is used. Other than for testing, why is *SecondaryStructure returned?

secondary_structure/dot_bracket.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	lenStructure := len(dotBracketStructure)
+	pc := &parseCompound{
+		length:             lenStructure,
+		pairTable:          pairTable,
+		annotatedStructure: make([]byte, lenStructure),
+	}
+
+	secondaryStructure := evaluateParseCompound(pc)
```
Koeng101, line 76: This evaluation is necessary to fill up parseCompound with data. Should evaluateParseCompound be a method of parseCompound then? That may make it more explicit that this call is having side effects (ie, modifying pc)

secondary_structure/seconday_structure.go
```diff
@@ -0,0 +1,251 @@
+package secondary_structure
```
Koeng101, line 1: This makes sense - is mainly just setting up the proper structs.

Koeng101 (commented):
Overall, looks good, just want to clean up `FromDotBracket`'s output, since it is one of the only public exported functions. If I could learn why SecondaryStructure is returned, that would be useful.

v-raja:
Thanks for the review @Koeng101. I've made a few changes:
- Removed the `dotBracketFromSecondaryStructure` func. I don't think people will actually use it. The code is available on [another branch](https://github.com/v-raja/poly/blob/dot_bracket_secondary_structures/secondary_structure/dot_bracket.go) in case someone wants it in the future.
- Updated the `MultiLoop` struct's `Substructures` field to be of type `[]interface{}` instead of `SecondaryStructure` (because multiloop substructures aren't really a `SecondaryStructure` of its own). 
- Better and more informative examples that help people understand what a `SecondaryStructure` struct contains. 

Ready for another review.

PR #178, opened 2021-07-16
Synthesis fixer update after `repackage` PR

v-raja:
This PR consists of the following:
- Merging `prime` branch into `synthesisFixer` after the `repackage` PR
- Refactor the synthesis stuff into its own subpackage and update func calls that changed due to the the `repackage` PR
- Update the data directory of the synthesis test files as path changed when moving to own subpackage
- Update the direct dependencies which resulted in the updating of `mroth/weightedrand` from `v0.2.1` to `v0.4.1`. In this update, the api of the func `weightedRand.NewChooser` was updated to include an err term. So `chooser()` was also updated to return this err term.


Koeng101:
I've updated this package to use https://gitlab.com/cznic/sqlite instead of sqlite3. This removes the need for CGO in poly, and in my brief testing there wasn't a huge performance impact. Before we commit, let me write a tutorial for this package.

Koeng101:
Added tutorial in example_test.go

PR #182, opened 2021-08-02
Added concurrent parsing feature for genbank files

Koeng101:
This adds simple concurrent parsing for genbank files. I added two functions:

1. ParseConcurrent: Parses a mutli-genbank file into a channel
2. ParseFlatConcurrent: Parses a flat-genbank file into a channel

I also added an MD5 hash into the Sequence struct. This will be used in allbase to see if genbank files should be updated.

### DIFF
--- a/io/genbank/genbank.go
+++ b/io/genbank/genbank.go
@@ -781,18 +781,14 @@ func ParseMulti(file []byte) []poly.Sequence {
 // the genbank ftp dumps. These files have 10 line headers, which are entirely
 // removed
 func ParseFlat(file []byte) []poly.Sequence {
-
-	gbk := string(file)
-
-	// This code removes the header, which is 10 lines long. This is inefficient
-	// and gets rid of the data in the header, which may be useful for some
-	// application. Header data is not needed to parse the Genbank files, though
-	gbkWithoutHeader := []byte(strings.Join(strings.Split(gbk, "\n")[10:], "\n"))
-
-	// Pass gbkWithoutHeader to ParseMulti, which should handle
-	// the rest of the parsing just fine
-	sequences := ParseMulti(gbkWithoutHeader)
-	return sequences
+	r := bytes.NewReader(file)
+	sequences := make(chan poly.Sequence, 1000) // A buffer is used so that the functions run as it is appending
+	go ParseFlatConcurrent(r, sequences)
+	var outputGenbanks []poly.Sequence
+	for sequence := range sequences {
+		outputGenbanks = append(outputGenbanks, sequence)
+	}
+	return outputGenbanks
 }
 
 // ReadMulti reads multiple genbank files from a single file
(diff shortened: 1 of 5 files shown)


io/genbank/genbank.go
```diff
@@ -35,6 +38,9 @@ func Parse(file []byte) poly.Sequence {
 	// Create sequence struct
 	sequence := poly.Sequence{}
 
+	// Add the SHA256 hash to sequence.SHA256
```
TimothyStiles, line 41: MD5 isn't SHA256

poly.go
```diff
@@ -103,6 +103,7 @@ type Sequence struct {
 	SequenceHashFunction string    `json:"hash_function"`
 	Sequence             string    `json:"sequence"`
 	Features             []Feature `json:"features"`
+	MD5                  [16]byte  `json:"hash"` //SHA256 checksum
```
TimothyStiles, line 106: MD5 is not SHA256

io/genbank/genbank.go
```diff
@@ -760,41 +766,29 @@ Genbank Flat specific IO related things begin here.
...
-	// and gets rid of the data in the header, which may be useful for some
-	// application. Header data is not needed to parse the Genbank files, though
-	gbkWithoutHeader := []byte(strings.Join(strings.Split(gbk, "\n")[10:], "\n"))
-
-	// Pass gbkWithoutHeader to ParseMulti, which should handle
-	// the rest of the parsing just fine
-	sequences := ParseMulti(gbkWithoutHeader)
-	return sequences
+	r := bytes.NewReader(file)
+	sequences := make(chan poly.Sequence, 1000) // A buffer is used so that the functions run as it is appending
```
TimothyStiles, line 785: Why 1000 channels?

io/genbank/genbank.go
```diff
@@ -760,41 +766,29 @@ Genbank Flat specific IO related things begin here.
 
 // ParseMulti parses multiple Genbank files in a byte array to multiple sequences
 func ParseMulti(file []byte) []poly.Sequence {
+	r := bytes.NewReader(file)
+	sequences := make(chan poly.Sequence, 1000) // A buffer is used so that the functions run as it is appending
```
TimothyStiles, line 770: Why 1000 channels?

io/genbank/genbank.go
```diff
@@ -35,6 +38,9 @@ func Parse(file []byte) poly.Sequence {
 	// Create sequence struct
 	sequence := poly.Sequence{}
 
+	// Add the SHA256 hash to sequence.SHA256
+	sequence.MD5 = md5.Sum(file)
```
TimothyStiles, line 42: Does Genbank publish their file hashes and how do we check against them?

Koeng101: No. This is for if you want to check if a genbank file has changed.

poly.go
```diff
@@ -103,6 +103,7 @@ type Sequence struct {
 	SequenceHashFunction string    `json:"hash_function"`
 	Sequence             string    `json:"sequence"`
 	Features             []Feature `json:"features"`
+	MD5                  [16]byte  `json:"hash"` //MD5 checksum
```
TimothyStiles, line 106: Considering that we already have sequence_hash and hash_function can we call this something like `GenbankCheckSum` in both the struct and json output @Koeng101?

Koeng101:
Resolved the naming issue (changed from MD5 to CheckSum). Now uses blake3. Additionally added this as a feature into the gff parser. 

Interestingly enough, GFF actually produces the exact file (as hashed) as its input, meaning it is far more reliable than Genbank.

TimothyStiles (approved):
Looks good to me. Approved for merge.

PR #184, opened 2021-08-07
Tutorials

Koeng101:
gob https://pkg.go.dev/encoding/gob has some nice tutorials in the pkg docs for their golang library. While we definitely need better documentation, I thought adding package documentation and very very simple tutorials would be a good start.

### DIFF
--- a/io/genbank/example_test.go
+++ b/io/genbank/example_test.go
@@ -2,11 +2,12 @@ package genbank_test
 
 import (
 	"fmt"
+
 	"github.com/TimothyStiles/poly/io/genbank"
 )
 
 // This example shows how to open a genbank file and search for a gene given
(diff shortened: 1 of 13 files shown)


io/genbank/example_test.go
```diff
@@ -0,0 +1,18 @@
+package genbank_test
+
+import (
+	"fmt"
+	"github.com/TimothyStiles/poly/io/genbank"
+)
+
+// This example shows how to open a genbank file and search for a gene given
+// its name. After finding it, notes about the particular gene is read.
```
TimothyStiles: *are read*

TimothyStiles (approved):
There could be more inline comments for the examples but as it stands this is better than what we have now so I'll merge it and we can add more later.

PR #185, opened 2021-08-12
Synthesis fix is not selecting the codon with the highest weight

isaacguerreir:
I created a test where we could see if the FixCds will select the codon with the highest weight. I altered the Picha Codon table to change the weight of the K amino acid, to have CGA instead of AGA as the codon with the highest weight.

I think this is happening by two possible explanations:

- They're choosing the codon by alphabetic order. One way to see this running is with the example that I made. This should generate a `synth.db`. You could see every step of the function importing `synth.db` using a online [sqlite viewer](https://inloop.github.io/sqlite-viewer/). 

- Some amino acids as K have only two possible synonymous codons (AAA and AAG). Because CdsFix is making `AND h.codon != cb.tocodon` it will be choosing the other one, and this could give us the impression that this is actually right (You could see this in the `ExampleFixCdsSimple` test).


Koeng101:
Interesting! This indeed fails. Let me check it out.

PR #186, opened 2021-08-12
Create a feature to remove global repetitions 

isaacguerreir:
The `GlobalRemoveRepeat` feature removes repetitions between CDS and an external sequence as host genome or plasmid to avoid homology recombination.

Repetitive genetic sequences when introduced in a host organism could lead to homologous recombination and the deletion of your sequence. According to this paper ["Automated design of thousands of nonrepetitive parts for engineering stable genetic systems"](https://www.nature.com/articles/s41587-020-0584-2.epdf?sharing_token=y4fo_gSeypnwZ4fnKOd6u9RgN0jAjWel9jnR3ZoTv0PpP8WpygJAKEZCXED_ZW9zc8juzH923QlERAqIKityDMvWODe8F5z99ZQpcYjQxxpYu-NKeObW6J1g778fvDV7gsVEEYAyqqP-kev-0neshA7In0mnUY_dM4kMyFWdTsQ%3D) a 20-bp repeat is enough to trigger detectable levels of recombination in E. coli. This could be important because CDS are great places to remove those types of repetitions without changing functionality.

I created to support this function a `GetKmerTabler` function in the transform subpackage.

While testing this `GlobalRemoveRepeat` I recognize an edge case when `FixCds` iterate 100 times without fixing all the problems. I made a fix that instead of returning an error, it checks if some change was made and returns those changes.


Koeng101:
Merging into synthesis fixer. I like this feature.

PR #187, opened 2021-08-17
Hairpin remover function for FixCds and test

isaacguerreir:
This feature is a hairpin remover based on the` AvoidHairpin` function from [DNA Chisel](https://github.com/Edinburgh-Genome-Foundry/DnaChisel/blob/db0b606ec993a509de9f3bb4d8acde9baadeef99/dnachisel/builtin_specifications/AvoidHairpins.py#L8) python package. By doing this is possible to minimize some impact of secondary structures in CDS. Some papers [(Design Parameters to Control Synthetic Gene Expression in Escherichia coli)](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0007002) indicate that secondary structure near the translational start site could be detrimental to protein production, this could make this less impactful after codon optimization.


isaacguerreir:
I had also added a simple fix to check if the sequence it's been passed through fixCds functions is actually a full codon sequence without interrupted (1-2bp) codons. Without this check and interrupted sequence could lead to SQLite panic logs.

PR #188, opened 2021-08-18
Create a list of all possible primers following a guideline

isaacguerreir:


primers/pcr/pcr.go
```diff
@@ -136,3 +144,83 @@ func Simulate(sequence string, targetTm float64, circular bool, primerList []str
...
+// of primers
+//
+// So to make this happen I will use the following approach:
+//
+// 1. I will take a sequence and generate two list of possibles primers using ranges passed by user as arguments
+// 2. I will use both lists to try to find pairs of primers that have the correct delta melting temperature between
+// a pair of primers
+// 3. Return all possible pair of primers
+
+func DesignAllPossiblePrimers(sequence string, minGcContent float64, maxGcContent float64, minMeltingTemp float64, maxMeltingTemp float64, minLength int, maxLength int, maxMeltingTempBetweenPrimers float64) []PairPrimers {
```
Koeng101, line 171: I think the API for primer design should be simplified here. Perhaps a rename to `DesignWithGcConstraint` would be more appropriate. Then, users could decide if they just care about tm or if they care about gc. Thoughts @tim?

Koeng101 (commented):
Few changes to API that I think would make things make more sense.

primers/pcr/pcr.go
```diff
@@ -136,3 +144,83 @@ func Simulate(sequence string, targetTm float64, circular bool, primerList []str
 	}
 	return pcrFragments
 }
+
+// Generally people that wants a pair of primers for a given sequence doesn't have a target Tm in mind.
```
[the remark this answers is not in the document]
isaacguerreir: I didn't know about this. In my experience in biotech projects we never use a specific Tm in mind but always some guidelines to follow. So most of the time if someone design a primer that are following the main points of some guideline that's ok for synthesis.

primers/pcr/pcr.go
```diff
@@ -136,3 +144,83 @@ func Simulate(sequence string, targetTm float64, circular bool, primerList []str
...
+// design of primers.
+
+// One of the problems of Primer Design Guidelines is they're slightly different from each other.
+// One example of these is Addgene Primer Design guidelines says the melting temperature should be between 50-60°C
+// Thermo Fisher presents another information, they say the desired melting temperature is actually between 65°C and 75°C
+// This another guideline from Premier Biosoft only says 'Primers with melting temperatures in the range of 52-58°C generally produce the best results'
+// Sources: https://www.addgene.org/protocols/primer-design/, https://www.thermofisher.com/blog/behindthebench/pcr-primer-design-tips/
+// http://www.premierbiosoft.com/tech_notes/PCR_Primer_Design.html
+
+// So to allow users to follow your own guidelines we will make a software that founds every pair of primers
```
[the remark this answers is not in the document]
isaacguerreir: Hahaha sorry!

primers/pcr/pcr.go
isaacguerreir: I was actually thinking about a implementations just like synthesis where we have functions that will will work as filter for primers and pair of primers, and the user could select exactly like the guideline they are following and ask the function the find the primers.

isaacguerreir: This could also allow users to create their own filter for this.

PR #189, opened 2021-08-19
Fix for genbank files exported from benchling that have a different index for slash in qualifiers

isaacguerreir:
I have found that GenBank files exported from Benchling actually have 29 spaces between the beginning of the line and the first slash that it's used to identify a qualifier. Without this correction, the 17 features inside the `benchling.gb` file will only return the first one. I created a test in the final part of the `genbank_test.go` for this.


Koeng101 (approved):
This PR makes sense and has a good test.

PR #198, opened 2021-09-07
Add fragmentation to synthesis directory

### LINKED ISSUE #84: Optimal fragmentation
I would like Poly to have a function to Fragment DNA samples that you wish to synthesize. Many DNAs that one would wish to synthesize are above the fragment threshold for synthesis providers (~1500bp for Twist), so this would be useful to them. In addition, you can synthesize more complex DNA fragments through clever splitting of Fragments (ie, split a homopolymer to get >8bp homopolymers in a synthesized fragment). 

I would like this function to be optimal using the data set in this paper - https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0238592 . You can imagine using a constraint solver to get the optimal overhangs for a given sequence, no matter the size. This is fundamentally useful for users who want to synthesize large fragments, like full biosynthetic pathways. It is also useful for users who want to efficiently create complex DNA from restrictive synthesis companies. 

I want a function that I can pass many sequences which then optimally fragments them for synthesis. 

**Additional context**
I'd like to be able to do "linking", ie, synthesizing many different genes on a single fragment, which then get divided using different restriction enzyme (ie, BbsI + BtgZI)

### LINKED ISSUE #196: SplitSet with more features
Part of the process of organism engineering is to make synthesizable parts, when necessary, to construct new sequences and plasmids that add functions to living things. One step that we found out very boring at Friendzymes is to make some parts synthesizable you have to split the original sequence into fragments, either because the original is too big to be synthesize, or because you have problems that could only be resolved by fragment the part. You not only have to make fragments for these parts, but you also have to find overhangs that actually are good and don't generate mismatches.

To create these optimized fragments people use the webtool from NEB called [SplitSet](https://ggtools.neb.com/splitset/run.cgi). They do this manually to produce this output. So @eyesmo asked me if we could make a version where we could automatize new heuristics to this process.

The result is this [repository](https://github.com/Open-Science-Global/SplitSet).

I have some things to resolve, but the main functionalities already work. We also found out that NEB doesn't select the best overhangs, but only finds one that is ok, so in this version of SplitSet we already have better results than the original SplitSet. And yeah, we're using the data they share in the paper and use it to select the best overhangs. (We checked this by using the [Ligase Fidelity Viewer](https://ggtools.neb.com/viewset/run.cgi), another webtool from NEB for this).

I also already integrate this package with IDT API, so we could check if the new fragments are synthesizable running the Complexity Score. If the score it's too high we try other fragmentation options to see if this improves the final selection of fragments.

We also integrate with the finder subpackage to find problematic parts in those sequences and prioritize that in the search of the perfect overhang for a given region we try to put a fragmentation in the middle of the problematic part to minimize even more the complexity score.

My question is more if this is something that will be interesting to have inside poly, or if it's better to have this in a different package. Let me know what you think about this.

Koeng101:
closes #84 and closes #196 

This adds in a fragmentation package to Poly and moves the synthesis package into the fix directory. 

In the end, our `synthesis` should have `fragment`, `fix`, and `optimize`.

Just need to increase the test coverage by a small amount.


Koeng101:
Will require upgrade from 0.13 -> 0.14 for Poly. Thought I should help with the repackage for synthesis associated stuff.

TimothyStiles:
Lmao codeclimate is actually making a good point. FixCds shouldn't be in a package called `fix` because you'd have to call `fix.FixCds` and it stutters doesn't really help tell you anything more about what you're calling at first glance.

Koeng101:
Fixed stutters

synthesis/fragment/example_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"fmt"
+	"github.com/TimothyStiles/poly/synthesis/fragment"
+)
+
+// This example shows how to use the fragmenter to fragment a gene in
+// preparation for synthesis.
+func Example_basic() {
+	lacZ := "ATGACCATGATTACGCCAAGCTTGCATGCCTGCAGGTCGACTCTAGAGGATCCCCGGGTACCGAGCTCGAATTCACTGGCCGTCGTTTTACAACGTCGTGACTGGGAAAACCCTGGCGTTACCCAACTTAATCGCCTTGCAGCACATCCCCCTTTCGCCAGCTGGCGTAATAGCGAAGAGGCCCGCACCGATCGCCCTTCCCAACAGTTGCGCAGCCTGAATGGCGAATGGCGCCTGATGCGGTATTTTCTCCTTACGCATCTGTGCGGTATTTCACACCGCATATGGTGCACTCTCAGTACAATCTGCTCTGATGCCGCATAG"
+	fragments, _, _ := fragment.Fragment(lacZ, 95, 105)
```
TimothyStiles, line 12: This stutters but code climate isn't yelling about it 🤔

TimothyStiles (approved):
There's a couple of things here and there name wise that could be a little better but this is a mighty fine PR and we can merge it!

PR #207, opened 2021-09-12
Synthesis fixer repeat needs love

Koeng101:
Added are a few complex genes with large repeats. Our repeat finding function needs love.


Koeng101:
Noted https://github.com/TimothyStiles/poly/blob/87dbfcae9320588144d9931fd620f3d50d9f9f1f/synthesis/synthesis_test.go#L207-L218

Will have to fix later.

PR #200, opened 2021-09-09
testing sql linter workflow.

TimothyStiles:

PR #208, opened 2021-09-15
Updated synthesisFixer to not use SQL.

Koeng101:
This push updates the synthesis fixer to not use SQL, but rather just use raw Golang. Should make troubleshooting easier.

Fixed 3 bugs in the example files:
* GGGCCC should fix to GGGCCA, not GGACCC, given its codon table
* repeat finder function should fix more of the repeat
* example_basic wasn't computing with a real file, so optimization table was set to zero. Now uses a real table

All other tests are passing.


Koeng101:
Also increases performance from 98053522 ns/op to 42113466 ns/op, a 2.3x speedup

Koeng101:
Closes #200

synthesis/synthesis.go
```diff
@@ -260,76 +178,60 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
+						atBiasMap[codon.Triplet] = append(atBiasMap[codon.Triplet], toCodon.Triplet)
 					case codonBias < toCodonBias:
-						tx.MustExec(`INSERT INTO codonbias(fromcodon, tocodon, gcbias) VALUES (?, ?, ?)`, codon.Triplet, toCodon.Triplet, "GC")
+						gcBiasMap[codon.Triplet] = append(gcBiasMap[codon.Triplet], toCodon.Triplet)
 					}
+					naBiasMap[codon.Triplet] = append(naBiasMap[codon.Triplet], toCodon.Triplet)
 				}
 			}
 		}
 		if aminoAcidTotal == 0 {
```
TimothyStiles, line 190: What is this conditional for?

synthesis/synthesis.go
```diff
@@ -260,76 +178,60 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
-			tx.MustExec(`INSERT INTO weights(codon, weight) VALUES (?,?)`, codon.Triplet, normalizedCodonWeight)
+			weightMap[codon.Triplet] = normalizedCodonWeight
 		}
 	}
 
-	err := tx.Commit()
-	if err != nil {
-		return sequence, []Change{}, err
+	// Build historical map
+	pos := 0
```
TimothyStiles, line 206: Should this be `position`?

synthesis/synthesis.go
```diff
@@ -260,76 +178,60 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
-	for i := 1; i < fixIterations; i++ {
+	getSequence := func(history map[int][]string) string {
+		var sequence string
+		for codonPosition := 0; codonPosition < len(history); codonPosition++ {
+			sequence = sequence + history[codonPosition][len(history[codonPosition])-1]
+		}
+		return sequence
+	}
+	var changes []Change
+	for step := 1; step < fixIterations; step++ {
```
TimothyStiles, line 224: What are the steps here? What is this for loop doing compared to the last?

synthesis/synthesis.go
```diff
@@ -260,76 +178,60 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
-					FROM   history AS h
-					       JOIN suggestedfix AS sf
-					         ON sf.id = h.suggestedfix
-					WHERE  h.suggestedfix IS NOT NULL
-					ORDER  BY h.step,
-					          h.pos 
-					`
-			_ = db.Select(&changes, getChangesSQL)
+			// Sort changes by step and position
+			sort.Slice(changes, func(i, j int) bool {
```
TimothyStiles, line 229: Could you name these lamba variables some other than `i` and `j`?

synthesis/synthesis.go
```diff
@@ -342,78 +244,66 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
-		               AND s.pos <= ?
-		               AND h.codon != cb.tocodon) AS t
-			LEFT JOIN (SELECT codon, pos FROM history WHERE pos >= ? AND pos <= ?) AS his ON t.codon = his.codon AND t.pos = his.pos
-					WHERE his.codon IS NULL
-					ORDER  BY t.step DESC, t.weight DESC) as t
-		GROUP  BY t.pos
-		LIMIT  ?; `
+			// For each suggestion, range between the values
+			var potentialChanges []Change
+			for positionSelector := suggestion.Start; positionSelector <= suggestion.End; positionSelector++ {
```
TimothyStiles, line 250: Better commenting for this control flow. Comments should make this easy to read at a glance.

synthesis/synthesis.go
```diff
@@ -342,78 +244,66 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
+					}
+				case "AT":
+					for _, potentialCodon := range atBiasMap[lastCodon] {
+						if _, ok := unavailableCodons[potentialCodon]; !ok {
+							potentialChanges = append(potentialChanges, Change{positionSelector, step, lastCodon, potentialCodon, suggestion.SuggestionType})
+						}
+					}
+				}
+			}
+			// Sort changes
```
TimothyStiles, line 278: What changes are we sorting?

TimothyStiles (changes requested):
This pull request is beautiful. In general it needs some better commenting so that people can skim through control flow and know what's going on. FixCDS should also have an in-depth comment that describes why this function and exists and how it works to do the thing.

Some variable name stuff as always but I liek dis @Koeng101.

synthesis/synthesis.go
Koeng101: Idiomatic go https://pkg.go.dev/sort#Slice

synthesis/synthesis.go
Koeng101: Cant divide by 0

Koeng101:
@TimothyStiles I added some more description of how to use into the docs. Fully describing the FixCds somewhat feels off to me to the basic user - most users probably shouldn't be using it (lambdas and shit). FixCdsSimple should be what they're using. Perhaps a better description for developers?

TimothyStiles:
@Koeng101 100% should have a better explanation for developers. I doubt biologists are going to go beyond FixCdsSimple but it's just not clear what sort of operations you're doing and why.

synthesis/synthesis.go
```diff
@@ -162,94 +154,23 @@ func findProblems(sequence string, problematicSequenceFuncs []func(string, chan
 // FixCds fixes a CDS given the CDS sequence, a codon table, a list of
 // functions to solve for, and a number of iterations to attempt fixing.
 // 10 iterations is a reasonable default for fixIterations.
-func FixCds(sqlitePath string, sequence string, codontable codon.Table, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup), fixIterations int) (string, []Change, error) {
+func FixCds(sequence string, codontable codon.Table, problematicSequenceFuncs []func(string, chan DnaSuggestion, *sync.WaitGroup), fixIterations int) (string, []Change, error) {
```
TimothyStiles, line 157: Should include some reference to FixCds simple for users that don't want to muck with details. Will save us time answering questions in long run.

synthesis/synthesis.go
```diff
@@ -162,94 +154,23 @@ func findProblems(sequence string, problematicSequenceFuncs []func(string, chan
...
+	// We have a historical map, a relative weight map, and a potential changes map.
+	historicalMap := make(map[int][]string)
+	weightMap := make(map[string]float64)
+	naBiasMap := make(map[string][]string)
+	gcBiasMap := make(map[string][]string)
+	atBiasMap := make(map[string][]string)
+
+	// Build historical maps and full amino acid weights
+	aminoAcidWeightTable := make(map[string]int)
 	for _, aminoAcid := range codontable.AminoAcids {
```
TimothyStiles, line 174: These triple nested for loops need line by line comments to explain what's happening.

synthesis/synthesis.go
```diff
@@ -260,76 +181,61 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
+						atBiasMap[codon.Triplet] = append(atBiasMap[codon.Triplet], toCodon.Triplet)
 					case codonBias < toCodonBias:
-						tx.MustExec(`INSERT INTO codonbias(fromcodon, tocodon, gcbias) VALUES (?, ?, ?)`, codon.Triplet, toCodon.Triplet, "GC")
+						gcBiasMap[codon.Triplet] = append(gcBiasMap[codon.Triplet], toCodon.Triplet)
 					}
+					naBiasMap[codon.Triplet] = append(naBiasMap[codon.Triplet], toCodon.Triplet)
 				}
 			}
 		}
+		// aminoAcidTotal can't equal zero or else division later will panic
```
TimothyStiles, line 193: What does aminoAcidTotal do though? Where is it used and how?

synthesis/synthesis.go
```diff
@@ -260,76 +181,61 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
-	for codonPosition := 0; codonPosition < len(sequence); codonPosition = codonPosition + codonLength {
-		codon := sequence[codonPosition : codonPosition+codonLength]
-		tx.MustExec(`INSERT INTO seq(pos) VALUES (?)`, pos)
-		tx.MustExec(`INSERT INTO history(pos, codon, step) VALUES (?, ?, 0)`, pos, codon)
-		pos++
+		aminoAcidWeightTable[aminoAcid.Letter] = aminoAcidTotal
 	}
 
+	// Build weight map
 	for _, aminoAcid := range codontable.AminoAcids {
```
TimothyStiles, line 201: "Cycle through every possible amino acid."

synthesis/synthesis.go
```diff
@@ -260,76 +181,61 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
 		for _, codon := range aminoAcid.Codons {
```
TimothyStiles, line 202: "For each codon related to said amino acid get the normalized codon weight ratio and add it to the weight map."

synthesis/synthesis.go
```diff
@@ -260,76 +181,61 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
 			normalizedCodonWeight := 100 * codonWeightRatio
-			tx.MustExec(`INSERT INTO weights(codon, weight) VALUES (?,?)`, codon.Triplet, normalizedCodonWeight)
+			weightMap[codon.Triplet] = normalizedCodonWeight
 		}
 	}
 
-	err := tx.Commit()
-	if err != nil {
-		return sequence, []Change{}, err
+	// Build historical map
```
TimothyStiles, line 209: What does the historical map do?

synthesis/synthesis.go
```diff
@@ -260,76 +181,61 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
 	}
 
 	// For a maximum of 100 iterations, see if we can do better. Usually sequences will be solved within 1-3 rounds,
 	// so 100 just effectively acts as the max cap for iterations. Once you get to 100, you pretty much know that
 	// we cannot fix the sequence.
-	for i := 1; i < fixIterations; i++ {
+	getSequence := func(history map[int][]string) string {
+		var sequence string
+		for codonPosition := 0; codonPosition < len(history); codonPosition++ {
+			sequence = sequence + history[codonPosition][len(history[codonPosition])-1]
```
TimothyStiles, line 223: Lot of messy indexing and inline arithmetic here. Find a more verbose way of expression that is easier to read.

synthesis/synthesis.go
```diff
@@ -342,78 +248,67 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
 			}
-			// First, let's insert the suggestions that we found using our problematicSequenceFuncs
-			_, err := db.Exec(`INSERT INTO suggestedfix(step, start, end, gcbias, quantityfixes, suggestiontype) VALUES (?, ?, ?, ?, ?, ?)`, i, suggestion.Start, suggestion.End, suggestion.Bias, suggestion.QuantityFixes, suggestion.SuggestionType)
-			if err != nil {
-				return sequence, []Change{}, fmt.Errorf("Failed to insert suggestedfix (does start+end point to valid parts of the amino acid sequence?). Got SQL error: %s", err)
+
+			// For each suggestion, get a list of potential changes that could fix the problem.
+			var potentialChanges []Change
+			for positionSelector := suggestion.Start; positionSelector <= suggestion.End; positionSelector++ {
+				lastCodon := historicalMap[positionSelector][len(historicalMap[positionSelector])-1]
```
TimothyStiles, line 255: Messy inline arithmetic. Expand.

synthesis/synthesis.go
```diff
@@ -342,78 +248,67 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
+				unavailableCodons := make(map[string]bool)
+				for _, codonSite := range historicalMap[positionSelector] {
+					unavailableCodons[codonSite] = true
+				}
+				// We will take new potential changes from the respective bias map, given the suggestion bias.
+				switch suggestion.Bias {
+				case "NA":
+					for _, potentialCodon := range naBiasMap[lastCodon] {
```
TimothyStiles, line 263: A lot of repeat control flow here. Consider making a callable helper function instead of copy and pasting.

synthesis/synthesis.go
```diff
@@ -342,78 +248,67 @@ func FixCds(sqlitePath string, sequence string, codontable codon.Table, problema
...
 
-	var changes []Change
-	_ = db.Select(&changes, `SELECT h.pos AS position, h.step AS step, (SELECT codon FROM history WHERE pos = h.pos AND step = h.step-1 LIMIT 1) AS codonfrom, h.codon AS codonto, sf.suggestiontype AS reason FROM history AS h JOIN suggestedfix AS sf ON sf.id = h.suggestedfix WHERE h.suggestedfix IS NOT NULL ORDER BY h.step, h.pos`)
+			// Make sure we have enough sorted changes after sorting/removal
+			if len(sortedChanges) < suggestion.QuantityFixes {
+				return sequence, []Change{}, fmt.Errorf("Too many fixes required. Number of potential fixes: %d , number of required fixes: %d", len(potentialChanges), suggestion.QuantityFixes)
+			}
+			targetChanges := sortedChanges[:suggestion.QuantityFixes]
 
+			// Update historical map, changes, and sequence
```
TimothyStiles, line 304: What does the historical map do?

TimothyStiles (changes requested):
More comments and refactoring of control flow necessary but this is starting to look pretty good.

synthesis/synthesis.go
Koeng101: I'll one up ya - how about a less verbose expression that is easier to read? Ie, use `i` instead of `codonPosition`.

This is essentially the golang equivalent of python's `my_list[-1]`, and tbh I think this "inline arithmetic" should be fairly clear.

synthesis/synthesis.go
Koeng101: It's not even arithmetic, It's just selecting the last element of a list

synthesis/synthesis.go
Koeng101: Yea, [it'd be a shame if some users](https://github.com/TimothyStiles/poly/blob/cfe26e05c3cac061452211ecccf14427230f397b/synthesis/synthesis.go#L10-L12) didn't read the docs. Will add.

synthesis/synthesis.go
Koeng101: > Where is it used and how
`/aminoAcidTotal` for vim users or `ctrl+f aminoAcidTotal` for normies

synthesis/synthesis.go
Koeng101: That's kinda what the code says already?

synthesis/synthesis.go
Koeng101: We're not training kids to read golang, they can see that we're ranging through amino acids and codons cause its right there, in fewer words than the comment. Will add the thing about adding to weight map.

synthesis/synthesis.go
Koeng101: Added

synthesis/synthesis.go
Koeng101: Fixed!

Koeng101:
Fixed @TimothyStiles !

Koeng101:
I removed the need for a magic number in the `findProblems` function. Probably out of scope for PR, but a nice addition I thought you might like.

TimothyStiles:
MERGED

TimothyStiles:
@Koeng101 think you could add an example test for FixCDS in a new PR?

PR #209, opened 2021-09-17
Fixed rare cloning bugs

Koeng101:
I fixed 2 rare cloning bugs:

1. In the case of an infinite cloning loop, return the last fragment of the loop. This should allow users to figure out which fragment is ineffective in their sequence. It also returns the "most likely" cloned seqeuence.
2. A panic is called if there are intercepting restriction enzyme sites. For example, `GAAGACATAATGGTCTTC` (two BbsI sites intercepting). This is now handled.


TimothyStiles:
I like this fix and these regression tests. MERGED.

PR #210, opened 2021-09-17
Synthesis fixer panic

Koeng101:
If the very end of a sequence is a restriction enzyme site, and the leftover arithmetic worked out in the right way, a panic could be called during synthesis fixing. This fixes the remove sequence function to make this not happen and adds a check within the synthesis fixer (future proofing for other synthesis fixing functions).

Included is a regression test.


synthesis/synthesis_test.go
```diff
@@ -226,3 +226,15 @@ func TestBadCodonTable(t *testing.T) {
...
+	_, _, _ = FixCdsSimple(gene, codonTable, []string{"GAAGAC", "GGTCTC", "GCGATG", "CGTCTC", "GCTCTTC", "CACCTGC", "CGTCTC"})
```
TimothyStiles, line 238: We're just assuming this test passes if we call the function and it doesn't panic correct? Sanity test for me because I'm tired and can't remember if this should work or not.

Koeng101: Yes, we're simply testing that the Go program doesn't go "AH FUCK".

PR #215, opened 2021-09-30
Get next overhang

Koeng101:
Added a function called `NextOverhangs` and `NextOverhang`. These can be used to generate the next optimal overhang for a list of overhangs. 

`NextOverhangs` is useful for optimal fragmentation of degenerate coded things (ie, if you're looking to split a gene up into different sections for synthesis and need optimal overhangs)

`NextOverhang` uses `NextOverhangs` and is useful when you're building a new list of optimal overhangs for particular applications. For example, I am using it for optimal overhangs for ligation of barcodes to a I-SceI site (this is also represented in the example for this function.

While Fragment currently fragments on the basis of sequence that is already there, these next 2 functions help if you're trying to fragment something which you have sequence control over.

100% test coverage.


Koeng101:
@TimothyStiles bump

TimothyStiles:
@Koeng101 checks are good. What's the code coverage?

synthesis/fragment/fragment.go
```diff
@@ -34,6 +35,58 @@ func SetEfficiency(overhangs []string) float64 {
...
+// list, along with their efficiencies. This can be used for more optimal
+// fragmentation of sequences with potential degeneracy.
+func NextOverhangs(currentOverhangs []string) ([]string, []float64) {
+	currentOverhangMap := make(map[string]bool)
+	for _, overhang := range currentOverhangs {
+		currentOverhangMap[overhang] = true
+	}
+	bases := []rune{'A', 'T', 'G', 'C'}
+	var overhangsToTest []string
+	for _, base1 := range bases {
```
TimothyStiles, line 48: okay I have to ask. What's with this super nested control flow? Can this be simplified?

TimothyStiles (changes requested):
@Koeng101

synthesis/fragment/fragment.go
Koeng101: Not sure if simplified, but could probably be smaller by using a recursive function. I think it'd be more complex tho, I think this is reasonable cause you can clearly see what it is doing.

Koeng101:
> @Koeng101 checks are good. What's the code coverage?

100% test coverage.

synthesis/fragment/fragment.go
Koeng101: Comment added.

PR #219, opened 2021-10-17
Refactoring docs to be more easily navigable.

TimothyStiles:
Hey y'all,

This PR is a work in progress where I'm refactoring Poly to make its docs more navigable. I've added package docs to every package and moved poly's core sequence struct to it's own dedicated package in the IO subpackage.

I'm going to write integrated tutorials for each subpackage as appropriate along with adding some doc.go and example_test files in top level packages.

If people could review this to help catch any mistakes that'd be great!

Thanks,
Tim

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -11,7 +11,7 @@ through a simple interface in Go.
 
 The reason why Go was choosen is that it's a highly popular
 language with a simple sytax and a large ecosystem of packages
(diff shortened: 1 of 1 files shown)


TimothyStiles:
I still need to write doc.go files for each top level subpackage along with examples_test files to top level packages but this is a good start and worlds better than what was here before.

PR #220, opened 2021-10-29
Added slow5 parser

Koeng101:
This PR will add a slow5 and blow5 format parser, from https://github.com/hasindu2008/slow5tools

This won't implement the fast5 -> slow5 functionality, but will be able to read and write those files for downstream use.


Koeng101:
Linter should be failing due to metadata being a lot. For coverage, we're going to have to add a ton of test files, which will be fun.

Koeng101:
https://pkg.go.dev/github.com/bmkessler/streamvbyte + https://pkg.go.dev/github.com/klauspost/compress/zstd should be useful for blow5 compression native to golang.

Koeng101:
@TimothyStiles Turns out I actually need this. I haven't implemented blow5, but slow5 is implemented here

PR #221, opened 2021-11-01
Genbank error handling

Koeng101:
This PR updates the Genbank parser to have both error handling and to function as a parser (ie, line by line, rather than simply passing most of the file into functions constantly).

Still a work in progress, but does parse a file successfully.

### DIFF
--- a/io/genbank2/genbank.go
+++ b/io/genbank2/genbank.go
@@ -1,0 +1,396 @@
+package genbank
+
+import (
+	"bufio"
+	"fmt"
+	"github.com/TimothyStiles/poly/io/poly"
+	"io"
+	"regexp"
+	"strings"
+)
+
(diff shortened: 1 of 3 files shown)


io/genbank2/genbank.go
```diff
@@ -0,0 +1,396 @@
...
+	"io"
+	"regexp"
+	"strings"
+)
+
+func Parse(r io.Reader) (poly.Sequence, error) {
+	scanner := bufio.NewScanner(r)
+
+	// Create meta struct
+	var meta = poly.Meta{}
```
rkrishnasanka, line 16: It'll be good have references for the file format and potentially any reference parsers

Koeng101:
Closed in favor of #244

Issue #201, opened 2021-09-11
Codon optimize does not check if amino acids are in table


Koeng101:
```
MGHHHHHHSKIFKOKNNYTOPKVFRDGCVPREIVYLPGNPAEIVLSTGERKIASCLHCPDAPCLSFKDHELQNSAFPEFPODQSAAVCATNAIAWDNENGIPVVDNNRCISCGICVDRCPVGAIYMIEEGIEINNNHIGDYLINLDEDDGDPSKIFGEKMYLLSEVKRSGKLKIETDEVITNIYNQLALIDLDAQFPNIFSRNLLMTLGTTCSIRRRGDVNVRMDAVLGPPGTNYGVMEVEFDSNALLDSPRNILDDLAVLSSRYGINYSDITPVIVSIAFPNTRSEYWRVINDINNVLQIKINSITIGALILLIWNLKKVNFNTDSFYADSECMEIRSSIANIIGDNPNISIGGILEVAK  }
panic: invalid argument to Intn

goroutine 22 [running]:
math/rand.(*Rand).Intn(0xc000656000, 0xc00006bb34)
        /usr/lib/go/src/math/rand/rand.go:168 +0x65
math/rand.Intn(...)
        /usr/lib/go/src/math/rand/rand.go:337
github.com/mroth/weightedrand.Chooser.Pick({{0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}, 0x0})
        /home/koeng/go/pkg/mod/github.com/mroth/weightedrand@v0.4.1/weightedrand.go:84 +0x4a
```

Basically, if you put in a protein sequence with invalid amino acids (in this case, O) optimize will panic instead of returning "hey you can't use that amino acid, idiot".

fialhopm:
Hi @Koeng101, would it be ok if I gave this a shot?

Koeng101:
> Hi @Koeng101, would it be ok if I gave this a shot?

That would be fantastic!

PR #228, opened 2021-11-29
Handle invalid amino acids in codon optimization

fialhopm:
This PR addresses issue #201 by updating `codon.Optimize` to gracefully handle protein sequences where at least one amino acid is missing from the codon table. A new custom error was added to enable the caller to find the offending amino acid.

Alternatively, `Optimize` could validate the protein sequence before building the codon chooser map. That would require a slightly bigger change to this function and would slow down the execution for valid inputs. On the upside, it could be beneficial to fail faster and to return the full set of invalid amino acids (as opposed to only the first one). Please let me know if that would be preferable.

### DIFF
--- a/synthesis/codon/codon.go
+++ b/synthesis/codon/codon.go
@@ -59,6 +59,14 @@ var errEmtpyCodonTable error = errors.New("empty codon table")
 var errEmtpyAminoAcidString error = errors.New("empty amino acid string")
 var errEmtpySequenceString error = errors.New("empty sequence string")
 
+type InvalidAminoAcidError struct {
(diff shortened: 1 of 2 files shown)


TimothyStiles:
@fialhopm somehow I didn't get a notification for this!

At first glance code looks really solid. Coverage dropped .6% in the codon package but that's easy to fix. I have family in town this weekend but on Monday I'll do a more proper review and we can get this merged!

@Koeng101 what do you think of this PR?

TimothyStiles:
@fialhopm looks like code coverage was a mistake on my end. Everything looks good. Merging!

PR #231, opened 2022-01-03
update .gitignore with new OS and EDITOR ignore templates

praveenRamalingam:
added linux/osx/windows and vim/vscode/emacs/jetbrains


praveenRamalingam:
ignored the vendor folder for golang, let me know if vendor should be checked in.

TimothyStiles (approved):

PR #214, opened 2021-09-30
Add GcContent function to barcodes

Koeng101:
This PR adds `CreateBarcodesGcRange` into `primers`. It gives users a more intuitive interface for generating barcodes that are within a given GC content, which is important when designing PCR primers and the like for barcoding sequences.

100% test coverage maintained.

PR #234, opened 2022-01-07
Revert "Add GcContent function to barcodes"

TimothyStiles:
Reverts TimothyStiles/poly#214


TimothyStiles:
Reverting previous commit because for some reason it fails tests when merged into prime branch.

PR #236, opened 2022-01-08
Added error checking to genbank 

Koeng101:
This adds error checking genbank to io.

There are a few things I need to figure out:

- Accession + Version + Definition formatting in build
- Locus comes up twice

I also need to figure out how to integrate concurrent reading (this will probably just be by modifying parse, even if that is a little annoying).

### DIFF
--- a/io/genbank/genbank.go
+++ b/io/genbank/genbank.go
@@ -13,17 +13,17 @@ package genbank
 import (
 	"bufio"
 	"bytes"
-	"compress/gzip"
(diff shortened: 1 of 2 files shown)


Koeng101:
@TimothyStiles This code is ready.

I removed checksums, multi-genbank, and flat genbank parsing. These I think can be added later after we clean up the code a little bit.

Koeng101:
Closed in favor of #244

PR #238, opened 2022-01-15
Example Primer Development Workflow

codercahol:
image

Getting the ball rolling on implementing an example primer workflow that satisfies the following in order to generate a large number of DNA primers:

- [x] consistent GC content
- [x] roughly the same melting temperature
- [ ] don't dimerize with themselves or other primers in the set
- [x] no shared subsequences >4bp with any other primer in the set
- [ ] does not bind to any DNA sequences from a reference (e.g. FreeGenes library and the genomes of E.coli, B.subtilis, S.cerevisiae, and P.pastoris)

### DIFF
--- a/primers/primers_test.go
+++ b/primers/primers_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/TimothyStiles/poly/checks"
 	"github.com/TimothyStiles/poly/transform"
 )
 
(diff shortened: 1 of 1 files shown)


codercahol:
I'm not sure what the condition for no dimerization ought to be; if it is that no two sequences are reverse complements of each other, then this condition is satisfied by the "no shared subsequences >4bp"

codercahol:
regarding: Doesn't Bind to Reference Sequences

hopefully is straightforward iteration over the genomes, but there's definitely room for a better(=faster) method since iterating over multiple Mbp genomes could easily add up to take a while

TimothyStiles:
Thanks for getting the ball rolling @codercahol! 

@eyesmo since you requested this can you help review it?

eyesmo:
Excellent. Preliminary thoughts:
1.

PR #244, opened 2022-03-11
Genbank error checking

Koeng101:
This is a genbank error checking file. As with the other branches, multi genbank parsing was removed. Parse, write, and read now defaults to outputting a list, though we don't have multi-genbank working quite yet.

I deleted `TestLocusParseRegression` because _it doesn't work_. These two lines are different, which the old parser **did not pick up**, but the newer parser picks up.
https://github.com/TimothyStiles/poly/blob/359da2c15a5b20c12b68d5eb25f8254dd67ee62a/data/puc19static.gbk#L43

https://github.com/TimothyStiles/poly/blob/359da2c15a5b20c12b68d5eb25f8254dd67ee62a/data/puc19.gbk#L43


io/genbank/genbank.go
```diff
@@ -821,7 +785,7 @@ func buildMetaString(name string, data string) string {
 		if index == 0 {
 			returnData = name + datum + "\n"
 		} else {
-			returnData += generateWhiteSpace(11) + datum + "\n"
+			returnData += generateWhiteSpace(12) + datum + "\n"
```
TimothyStiles, line 788: why the change to 12?

io/genbank/genbank.go
```diff
@@ -803,6 +766,7 @@ func parseLocation(locationString string) Location {
 		location = location.SubLocations[0]
 	}
 
+	location.GbkLocationString = locationString
```
TimothyStiles, line 769: This assignment doesn't do anything?

TimothyStiles:
Biggest concern is that the original Genbank/MultiGenbank paradigm has been overwritten where now MultiGenbank is now just being called Genbank. Can we have both?

io/genbank/genbank.go
```diff
@@ -72,7 +71,6 @@ type Feature struct {
 
 // Reference holds information for one reference in a Meta struct.
 type Reference struct {
-	Index   string `json:"index"`
```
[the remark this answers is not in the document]
Koeng101: Index was redundant I believe, literally coded twice.

io/genbank/genbank.go
Koeng101: Correct, removed. Sorry about that.

io/genbank/genbank.go
Koeng101: Previously this behavior was just broken. See:

https://github.com/TimothyStiles/poly/blob/7f133c87716de08b873b44d5a405688da1ddebb4/data/puc19static.gbk#L10-L11

Koeng101: Should be aligned

Koeng101:
> Biggest concern is that the original Genbank/MultiGenbank paradigm has been overwritten where now MultiGenbank is now just being called Genbank. Can we have both?

Yes - Genbank is just MultiGenbank of list size 1 (in fact, the parser only does single Genbank right now, gotta fix it up to do multi. Lemme try that out). Much easier than having separate functions I think. 

Removed redundancy in latest push and deleted puc19static - which was broken in multiple ways.

TimothyStiles:
@Koeng101 don't forget to merge upstream. I'm pretty sure I deleted that static puc19 file in my last merged PR.

TimothyStiles:
> > Biggest concern is that the original Genbank/MultiGenbank paradigm has been overwritten where now MultiGenbank is now just being called Genbank. Can we have both?
> 
> Yes - Genbank is just MultiGenbank of list size 1 (in fact, the parser only does single Genbank right now, gotta fix it up to do multi. Lemme try that out). Much easier than having separate functions I think.

@Koeng101 I think separate functions would be better for clarity but we can debate later. It'd be super easy to add a variadic argument to a multigenbank parser to limit number of genbanks parsed and just call that from our "single" genbank parser.

Koeng101:
> @Koeng101 don't forget to merge upstream. I'm pretty sure I deleted that static puc19 file in my last merged PR.

Yes, it is merged upstream, and no, it wasn't deleted because it was used for testing.

> I think separate functions would be better for clarity but we can debate later

Ye that is fair

Koeng101:
@TimothyStiles How is this looking? I recall I removed the multi-genbank functions, but is there anything else?

io/genbank/genbank.go
```diff
@@ -147,221 +145,290 @@ func getFeatureSequence(feature Feature, location Location) (string, error) {
...
 	}
+	outputMetadata = outputMetadata[:len(outputMetadata)-1] // Remove trailing whitespace
+	return outputMetadata
+}
 
-	// start writing features section.
-	gbkString.WriteString("FEATURES             Location/Qualifiers\n")
-	for _, feature := range sequence.Features {
-		gbkString.WriteString(BuildFeatureString(feature))
+func parseReferences(metadataData []string) (Reference, error) {
```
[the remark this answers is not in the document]
TimothyStiles: will remain to make it easier to batch rename variables when compared to metadataTag.

io/genbank/genbank.go
```diff
@@ -147,221 +145,290 @@ func getFeatureSequence(feature Feature, location Location) (string, error) {
...
+	case "TITLE":
+		reference.Title = referenceValue
+	case "JOURNAL":
+		reference.Journal = referenceValue
+	case "PUBMED":
+		reference.PubMed = referenceValue
+	case "REMARK":
+		reference.Remark = referenceValue
+	default:
+		return fmt.Errorf("ReferenceKey not in [AUTHORS, TITLE, JOURNAL, PUBMED, REMARK]. Got: %s", referenceKey)
```
TimothyStiles, line 414: any test case for this?

io/genbank/genbank.go
```diff
@@ -898,114 +861,3 @@ func generateWhiteSpace(length int) string {
 GBK specific IO related things end here.
 
 ******************************************************************************/
-
-/******************************************************************************
-
```
TimothyStiles, line 903: @Koeng101 did you mean to deprecate the below section?

TimothyStiles (changes requested):

io/genbank/genbank.go
```diff
@@ -146,222 +144,314 @@ func getFeatureSequence(feature Feature, location Location) (string, error) {
...
+	var sequenceBuilder strings.Builder
 
-	accessionString := buildMetaString("ACCESSION", sequence.Meta.Accession)
-	gbkString.WriteString(accessionString)
+	// Sequence breaking
+	parseStep := "metadata" // we can be parsing metadata, features, or sequence
 
-	versionString := buildMetaString("VERSION", sequence.Meta.Version)
-	gbkString.WriteString(versionString)
+	genbankStarted := false
```
TimothyStiles, line 203: @Koeng101 was this supposed to be used somewhere? The flag is set to true but it doesn't seem to ever be checked.

io/genbank/genbank.go
```diff
@@ -146,222 +144,548 @@ func getFeatureSequence(feature Feature, location Location) (string, error) {
...
+		gbkString.WriteString(sourceString)
+
+		organismString := buildMetaString("  ORGANISM", sequence.Meta.Organism)
+		gbkString.WriteString(organismString)
+
+		var taxonomyString strings.Builder
+		for i, taxonomyData := range sequence.Meta.Taxonomy {
+			taxonomyString.WriteString(taxonomyData)
+			if len(taxonomyData) == i+1 {
+				taxonomyString.WriteString(".")
```
TimothyStiles, line 256: @koeng this logic is covered in tests. What does it do exactly?

io/genbank/genbank.go
```diff
@@ -146,222 +144,551 @@ func getFeatureSequence(feature Feature, location Location) (string, error) {
...
+			parameters.newLocation = false
+
+			// First, let's check if we have a quote active (basically, if a attribute is using multiple lines)
+			if parameters.quoteActive {
+				trimmedLine := strings.TrimSpace(line)
+				if len(trimmedLine) < 1 {
+					continue
+				}
+				// If the trimmed line ends, append to the attribute and set parameters.quoteActive set to false
+				if trimmedLine[len(trimmedLine)-1] == '"' {
```
TimothyStiles, line 546: this won't work with all labels. Labels in t4_intron.gb don't have quotes surrounding label string.

TimothyStiles: attention @Koeng101 can you work on this bug? Essentially you can't rely on quotations to show you the end of an attribute because some one line attributes don't use any quotations like this:

```
 gene            1..243
                 /gene="frd.1"
                 /locus_tag="T4p240"
                 /label=frd.1  <--- offending line
                 /db_xref="GeneID:1258741"

```

TimothyStiles:
@Koeng101 found an issue with the genbank parser where it somehow strips out complements from the gbk string while parsing. Any idea on what's going on here?
Screen Shot 2022-06-18 at 4 46 51 AM
Screen Shot 2022-06-18 at 4 45 58 AM

TimothyStiles:
Also need to fix qualifier line breaks but that should be easy enoug
Screen Shot 2022-06-18 at 4 51 23 AM
h

PR #246, opened 2022-03-27
Increase the coverage of the codon package

fergloragain:
In this PR, I've made a few stylistic tweaks to the code, e.g. multiple var declarations moved into a single var block. As well as this, I've made a few corrections to spelling mistakes and typos in the codon.go file. 

The main change here is, the introduction of the `Table` interface type. This type is now used in a number of function declarations, in place of the old `Table` struct, which has been renamed to `codonTable`. The idea is that, if we pass the `Table` interface type into functions, then our tests can implement the `Table` interface with mock table types, which in turn allows us to orchestrate error scenarios.

I've also introduced the `newChooserFn` var, which we can substitute out in our tests. This change, along with the new `Table` interface, means we can achieve 100% coverage in this package.

I understand that the changes here are a bit more involved than my previous PR, so please feel free to ask any questions! I hope this approach doesn't go against any of the coding standards in this project :+1:


Koeng101 (approved):
Perfect kind of PR we need. I wholeheartedly approve.

Koeng101:
PS: Don't worry too much about codeclimate. Tim mostly just uses it for checks, but it sometimes makes absurd requests.

synthesis/codon/codon.go
```diff
@@ -193,27 +208,31 @@ func getCodonFrequency(sequence string) map[string]int {
...
 }
 
-// chooser is a Table method to convert a codon table to a chooser
-func (codonTable Table) chooser() (map[string]weightedRand.Chooser, error) {
+func (t codonTable) IsEmpty() bool {
+	return len(t.StartCodons) == 0 && len(t.StopCodons) == 0 && len(t.AminoAcids) == 0
+}
+
+// Chooser is a codonTable method to convert a codon table to a chooser
+func (t codonTable) Chooser() (map[string]weightedRand.Chooser, error) {
```
TimothyStiles, line 216: can `t` become `table`?

synthesis/codon/codon.go
```diff
@@ -232,17 +251,29 @@ func (codonTable Table) chooser() (map[string]weightedRand.Chooser, error) {
...
-	for _, aminoAcid := range codonTable.AminoAcids {
+	for _, aminoAcid := range t.AminoAcids {
 		for _, codon := range aminoAcid.Codons {
 			translationMap[codon.Triplet] = aminoAcid.Letter
 		}
 	}
 	return translationMap
 }
 
+func (t codonTable) GetStartCodons() []string {
```
TimothyStiles, line 265: `t` to `table` no single letter variables.

TimothyStiles:
This PR is also excellent @fergloragain! It looks like the only user facing stuff that changes are the Table's interface method? Is there anything else?

Can you explain more about how mocking will work testing wise? I've generally tried to stay away from interfaces until recently because it feels like a lot of boilerplate. 

`newChooserFn` looks like a good idea since we'll want to swap out the third party chooser we're using now anyways.

fergloragain:
> This PR is also excellent @fergloragain! It looks like the only user facing stuff that changes are the Table's interface method? Is there anything else?
> 
> Can you explain more about how mocking will work testing wise? I've generally tried to stay away from interfaces until recently because it feels like a lot of boilerplate.
> 
> `newChooserFn` looks like a good idea since we'll want to swap out the third party chooser we're using now anyways.

Yeah I think the user facing changes are limited to the new interface. 

So at this stage I decided to provide an interface that captured all of the functions exposed by the `codonTable` type. This was because I wanted a way of mocking the `Chooser()` functionality, to get the error flow for that function covered by our tests. Given that this is the only function I wanted to mock, we really only need a have an interface with a single `Chooser` method defined. However, I do think it's useful to be able to mock every function of a type, so that we can simulate error flows in the future if needed. I get that the extra boiler plate that comes with needing to implement all the functions on a mock type when we don't need them, so I've updated the code so that the `mockTable` type embeds a `codonTable` type. That way, we only need to mock the functions that the function we're testing needs. Does that answer your question?

TimothyStiles:
> That way, we only need to mock the functions that the function we're testing needs. Does that answer your question?

More of less yes! Could use a lesson on the finer point of mocking but I really like this PR.

Thanks @fergloragain. MERGED.

PR #270, opened 2022-09-25
Install Go Analysis Tools in Gitpod workspace

lgruen:
Without this, VS Code prompts to install `gopls`, `dlv`, etc. before setting a breakpoint + debugging a tutorial works.

I'm not sure if there's a more elegant way to run `Go: Install/Update Tools` (https://github.com/golang/vscode-go/wiki/tools) in the Command Palette instead.

Also simplified the check for `$tutorial`, as otherwise this line causes a syntax error when the variable isn't set at all.


TimothyStiles:
This works great! Thanks @lgruen! Merged.

PR #271, opened 2022-09-26
Open tutorial README in Gitpod

lgruen:
Only open the tutorial README file instead of restricting the view to the `tutorials` directory:

- This avoids reloading the workspace, which at least in the browser can cause a second confirmation prompt.
- It's nice to be able to browse the full code in the Explorer bar and jump to the right file from the tutorial. That's also why installing `github.com/rogpeppe/godef` is helpful, to jump directly to function definitions.


TimothyStiles:
Good points @lgruen! Merged.

PR #273, opened 2022-10-15
Move enzymeMap to package level, compile regular expressions only once

matiasinsaurralde:
I think that this might be useful and perform better if you're calling `getBaseRestrictionEnzymes` multiple times in the same function. Before this change, `getBaseRestrictionEnzymes` would compile the same regular expressions over and over.

### DIFF
--- a/clone/clone.go
+++ b/clone/clone.go
@@ -86,23 +86,20 @@ type Enzyme struct {
 	RecognitionSite string
 }
 
+// Eventually, we want to get the data for this map from ftp://ftp.neb.com/pub/rebase
+var enzymeMap = map[string]Enzyme{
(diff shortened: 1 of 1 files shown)


TimothyStiles:
@matiasinsaurralde it looks like @Koeng101 defined `enzymeMap` as a function so that it'd behave as a constant. 

https://qvault.io/2019/10/21/golang-constant-maps-slices/

Do you remember the reasoning behind this @Koeng101?

TimothyStiles:
Thanks for the PR @matiasinsaurralde! It looks like @Koeng101 defined enzymeMap as a function so that it'd behave as a constant but maybe it's no longer necessary that it does.

https://qvault.io/2019/10/21/golang-constant-maps-slices/

Do you remember the reasoning behind this @Koeng101?

Koeng101 (approved):
LGTM

Koeng101:
The reason was I just wanted an easy way to see default enzymes. However, the reasoning to put it behind a function makes sense to me.

PR #282, opened 2022-11-27
Fix README installation method

soypat:
New go modules requires version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Poly is a Go package for engineering organisms.
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


TimothyStiles:
> Hey @soypat
> 
> Its not a main package, so go install won't work...
> 
> maybe you mean go get ?

Thanks for pointing this out @gedw99! Just fixed it.

https://github.com/TimothyStiles/poly/commit/f96ebe0cdf4a237458d0854ea283cce434649d7b

PR #283, opened 2022-11-27
transform: performance like it matters

soypat:
Lot of noise in this PR, sorry about that. What matters is that
* I'm using a 256byte table for complement calculation instead of map
* I'm avoiding heap allocations where possible by using a byte slice and unsafe conversion.

```go
// ReverseComplement takes the reverse complement of a sequence.
// ReverseComplement expects ASCII input.
func ReverseComplement(sequence string) string {
    n := len(sequence)
    newSeq := make([]byte, n)
    for i := 0; i < n; i++ {
        newSeq[i] = complementTable[sequence[n-i-1]]
    }
    // This is how strings.Builder works with the String() method. If Mr. Go says it's safe...
    return *(*string)(unsafe.Pointer(&newSeq))
}
```

### DIFF
--- a/transform/examples_test.go
+++ b/transform/examples_test.go
@@ -2,7 +2,6 @@ package transform_test
 
 import (
 	"fmt"
-	"testing"
 
 	"github.com/TimothyStiles/poly/transform"
 )
@@ -29,51 +28,4 @@ func ExampleReverse() {
 	fmt.Println(reverse)
 
 	// Output: ACATTAG
-}
-func TestA(t *testing.T) {
-	for k, v := range complementBaseRuneMap2 {
-		got := transform.ComplementBase(k)
-		if v != got {
-			t.Errorf("%q: %q %q", k, v, got)
-		}
-		gotInverse := transform.ComplementBase(got)
-		if gotInverse != k {
-			t.Errorf("%q: %q %q", got, k, gotInverse)
-		}
-	}
-}
-
-var complementBaseRuneMap2 = map[rune]rune{
-	'A': 'T',
-	'B': 'V',
-	'C': 'G',
-	'D': 'H',
-	'G': 'C',
-	'H': 'D',
-	'K': 'M',
-	'M': 'K',
-	'N': 'N',
-	'R': 'Y',
-	'S': 'S',
-	'T': 'A',
-	'U': 'A',
-	'V': 'B',
-	'W': 'W',
-	'Y': 'R',
-	'a': 't',
-	'b': 'v',
-	'c': 'g',
-	'd': 'h',
-	'g': 'a',
-	'h': 'd',
-	'k': 'm',
-	'm': 'k',
-	'n': 'n',
-	'r': 'y',
-	's': 's',
-	't': 'a',
-	'u': 'a',
-	'v': 'b',
-	'w': 'w',
-	'y': 'r',
 }
--- a/transform/examples_test.go
+++ b/transform/examples_test.go
@@ -1,0 +1,79 @@
+package transform_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/TimothyStiles/poly/transform"
+)
+
+func ExampleReverseComplement() {
+	sequence := "GATTACA"
+	reverseComplement := transform.ReverseComplement(sequence)
+	fmt.Println(reverseComplement)
+
+	// Output: TGTAATC
+}
+
+func ExampleComplement() {
+	sequence := "GATTACA"
+	complement := transform.Complement(sequence)
+	fmt.Println(complement)
+
+	// Output: CTAATGT
+}
+
+func ExampleReverse() {
+	sequence := "GATTACA"
+	reverse := transform.Reverse(sequence)
+	fmt.Println(reverse)
+
+	// Output: ACATTAG
+}
+func TestA(t *testing.T) {
+	for k, v := range complementBaseRuneMap2 {
+		got := transform.ComplementBase(k)
+		if v != got {
+			t.Errorf("%q: %q %q", k, v, got)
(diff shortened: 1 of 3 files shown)


soypat:
Test is failing because I included a test for the complement table and it seems it was doing something I did not expect. 'U's complement is  'A'?

transform/examples_test.go
```diff
@@ -0,0 +1,31 @@
+package transform_test
+
+import (
+	"fmt"
```
Koeng101, line 4: Splitting into a separate file for examples + tests is great

transform/transform.go
```diff
@@ -12,74 +12,83 @@ ReverseComplement takes the reverse complement of a sequence.
...
-	for _, base := range complementString {
-		length--
-		newString[length] = base
+	n := len(sequence)
+	newSeq := make([]byte, n)
+	for i := 0; i < n; i++ {
+		newSeq[i] = complementTable[sequence[n-i-1]]
 	}
-	return string(newString)
+	// This is how strings.Builder works with the String() method. If Mr. Go says it's safe...
```
Koeng101, line 25: Useful comment

transform/transform.go
```diff
@@ -12,74 +12,83 @@ ReverseComplement takes the reverse complement of a sequence.
...
 func ReverseComplement(sequence string) string {
-	complementString := strings.Map(ComplementBase, sequence)
-	length := len(complementString)
-	newString := make([]rune, length)
-	for _, base := range complementString {
-		length--
-		newString[length] = base
+	n := len(sequence)
+	newSeq := make([]byte, n)
+	for i := 0; i < n; i++ {
```
Koeng101, line 22: @TimothyStiles doesn't let any single letter variables in. Could you change the i and n to something longer / more descriptive? (like above)

transform/transform.go
```diff
@@ -12,74 +12,83 @@ ReverseComplement takes the reverse complement of a sequence.
...
 // Reverse takes the reverse of a sequence.
 func Reverse(sequence string) string {
-	length := len(sequence)
-	newString := make([]rune, length)
-	for _, base := range sequence {
-		length--
-		newString[length] = base
+	n := len(sequence)
+	newSeq := make([]byte, n)
+	for i := 0; i < n; i++ {
```
Koeng101, line 44: Same as above. Single letter -> longer

transform/transform_test.go
```diff
@@ -1,31 +1,97 @@
...
+}
 
-	// Output: ACATTAG
+func BenchmarkReverse(b *testing.B) {
+	rng := rand.New(rand.NewSource(1))
+	seq := randomSequence(rng)
+	for i := 0; i < b.N; i++ {
+		seq = Reverse(seq)
+	}
+	_ = seq
```
Koeng101, line 96: Is there a benchmark comparison to old method? Would like to see!

Koeng101 (approved):

Koeng101:
A few things to change (single letter variables) - mostly interested in benchmarks. How much faster is it?

TimothyStiles:
I can fix variables and stuff but I'm also interested in benchmarks! Would be happy to start on a real benchmark suite for Poly @soypat.

soypat:
@Koeng101  @TimothyStiles I'll fix the variable names and add better documentation today- but I'd like to note that `i` as a generic counter in small scopes has been shown to be as readable or even more readable than a full length name. See [Beniamini et al.](https://www.cs.huji.ac.il/w~feit/papers/SingleLetter17ICPC.pdf).

It'd be awesome to have a benchmark suite like you say to get a real taste of how performance changes affects those using the library :smile:, more than happy to help out however I can. Been looking at refactoring the `fasta` package :eyes:.

<details><summary>Benchmarks for 32byte string input (click)</summary>

#### Before
```
goos: linux
goarch: amd64
pkg: github.com/TimothyStiles/poly/transform
cpu: Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
BenchmarkReverseComplement-8   	 1639549	       714.7 ns/op	     224 B/op	       3 allocs/op
BenchmarkComplement-8          	 3199111	       383.4 ns/op	      48 B/op	       1 allocs/op
BenchmarkReverse-8             	 4345659	       325.0 ns/op	     176 B/op	       2 allocs/op
PASS
```
#### After
```
BenchmarkReverseComplement-8   	18890924	        60.14 ns/op	      32 B/op	       1 allocs/op
BenchmarkComplement-8          	25396846	        55.45 ns/op	      32 B/op	       1 allocs/op
BenchmarkReverse-8             	19949305	        55.67 ns/op	      32 B/op	       1 allocs/op
```

</details>

transform/transform_test.go
```diff
@@ -60,6 +60,13 @@ func TestReverseComplement(t *testing.T) {
 	}
 }
 
+func TestComplementBaseError(t *testing.T) {
+	complementBase := ComplementBase('!')
+	if complementBase != ' ' {
```
TimothyStiles, line 65: @Koeng101 if complement base can't find a complement for the given nucleotide what should it return? N?

transform/transform_test.go
```diff
@@ -3,97 +3,86 @@ package transform
...
+	for _, testSequence := range evenAndOddLengthSequences {
+		got := ReverseComplement(testSequence)
+		expect := Reverse(Complement(testSequence))
 		if got != expect {
 			t.Errorf("mismatch with individual Reverse and Complement call:\n%q\n%q", got, expect)
 		}
 	}
 }
 
-func TestComplementBase(t *testing.T) {
```
soypat: Why was this test removed?

TimothyStiles: Added it back.

PR #284, opened 2022-11-29
fasta rework

soypat:
So this is **NOT** intended as a performance boost change. This is above all an API rework for ease of use and clarity.

I have implemented a fasta.Parser type. This type is intended to be the base to parse all fasta data and replace all other implementations.

#### Benefits of `Parser.ParseNext()` when compared to `XConcurrent` functions:
* Provides the bare minimum API signature to parse a single fasta genome
* It is a simple and easy to understand API.
* No ambiguities on API use.
* Takes a `io.Reader` for user liberty to use with any API or stream.
* User has more control over how many fasta genomes they want to read
* (EDIT) Also errors! ParseNext can return a very useful error with the line number on which parsing stopped 
```
goos: linux
goarch: amd64
pkg: github.com/TimothyStiles/poly/io/fasta
cpu: Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
BenchmarkFastaLegacy-8   	      55	  19968425 ns/op	17326897 B/op	  158406 allocs/op
BenchmarkParser-8        	     118	   9643441 ns/op	16795447 B/op	   58150 allocs/op
```


io/fasta/fasta.go
```diff
@@ -76,6 +80,143 @@ func Parse(r io.Reader) ([]Fasta, error) {
...
+		}
+		// If we got to this point we are currently inside of the fasta
+		// sequence contents. We append line to what we found of sequence so far.
+		sequence = append(sequence, line...)
+	}
+	// Parsing ended. Check for inconsistencies.
+	if lookingForName {
+		return Fasta{}, totalRead, fmt.Errorf("did not find fasta start '>', got to line %d", p.line)
+	}
+	if lookingForName && len(sequence) == 0 {
```
soypat, line 204: Oh snap, bug! This line should read 

```go
if !lookingForName && len(sequence) == 0 {
    // We found a fasta name but no sequence to go with it.
```

io/fasta/fasta.go
```diff
@@ -76,6 +80,150 @@ func Parse(r io.Reader) ([]Fasta, error) {
...
+type Parser struct {
+	// rd keeps state of current reader.
+	rd   bufio.Reader
+	line uint
+}
+
+// NewParser returns a Parser that uses r as the source
+// from which to parse fasta formatted sequences.
+func NewParser(r io.Reader) *Parser {
+	// What's the largest line size a fasta can be?
```
TimothyStiles, line 96: I believe there's no upper limit for fasta string length but some do limit to a certain number of characters per line. 70-80 appears to be common.

https://en.wikipedia.org/wiki/FASTA_format#Original_format_&_overview

soypat: Buffer could be **considerably** smaller then, heh.

tutorials/000_introduction_test.go
```diff
@@ -22,7 +22,7 @@ to ship a set of real tutorials to get people started with this special
 little library but I just couldn't find the right medium to ship a stable,
 deployable, and accessible tutorial series until today.
 
-This book is built into Poly's source code itself. Each tutorial a series
+This book is built into Poly's source code itself. Each tutorial is a series
```
TimothyStiles, line 25: I believe this was also in the last push you submitted @soypat? Do you know what's going on here?

io/fasta/fasta_test.go
```diff
@@ -11,6 +11,31 @@ import (
...
 
+func BenchmarkFastaLegacy(b *testing.B) {
+	var fastas []Fasta
+	for i := 0; i < b.N; i++ {
+		fastas, _ = Read("data/uniprot_1mb_test.fasta")
+	}
+	_ = fastas
+}
+
+func BenchmarkParser(b *testing.B) {
```
TimothyStiles, line 22: this benchmark errors out and doesn't hit any code.
Screen Shot 2022-12-01 at 8 34 13 PM

TimothyStiles:
@soypat the benchmark needs some fixing but this is looking really good! Is your plan to eventually deprecate the legacy fasta parsers as well?

io/fasta/fasta_test.go
soypat: Yeah, these benchmarks depended on unzipping the gz fasta, my bad. Fixing now!

tutorials/000_introduction_test.go
soypat: Git mystifies me. This tends to happen to me sometimes. I can try squashing commits, or that step can be done via github. As you'd prefer

soypat:
@TimothyStiles I'd really like to see those "Concurrent" parsers go. Maybe `Read` can be renamed as `ReadFile`, but that'd be gravy. Parse is OK as shorthand for creating a new parser and reading all.

soypat:
@TimothyStiles  So I have changed the following things:
* I initialize uniprot 1MB fasta as a raw format string in TestMain. This will avoid some noisiness of file manipulation in benchmarks
* I have added an argument to NewParser: maxLineSize, in bytes. This is to give users some freedom of how much heap they want to use for the parser. 
* Added error handling during benchmarks
* Better error handling for the non-EOF case during parsing

tutorials/000_introduction_test.go
TimothyStiles: Whatever is easiest! @soypat

TimothyStiles:
@soypat recent additions are looking great. Having a shorthand like `Parse` would be excellent.

`Reset`, `ParseAll` and `ParseN` need to be tested and it looks there are a couple of corner cases in `ParseNext` that also need to be tested.
Screen Shot 2022-12-02 at 1 16 24 PM
Screen Shot 2022-12-02 at 1 16 32 PM

soypat:
@TimothyStiles Alright, looks much better. Fixed various bugs and test coverage is up to 95%

io/fasta/fasta.go
```diff
@@ -76,6 +80,176 @@ func Parse(r io.Reader) ([]Fasta, error) {
...
+		fasta, _, err := p.ParseNext()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				err = nil // EOF not treated as parsing error.
+			}
+			return fastas, err
+		}
+		fastas = append(fastas, fasta)
+	}
+	return fastas, nil
```
TimothyStiles, line 121: success case doesn't seem to be covered by tests?

TimothyStiles:
@soypat I just rewrote `fasta.Parse` as a wrapper around `fasta.ParseAll` and had a couple of tests fail.

`ExampleWrite` fails and doesn't return anything when trying to read what was written out. Couple of things are happening here.

1. New parser is stripping the `*` character which denotes a stop codon. Since it's technically not a amino acid it doesn't have a letter but is important to denote.
2. It looks like there's a fatal difference between what the parser can read and what the writer outputs.

Screen Shot 2022-12-06 at 1 44 25 PM
Screen Shot 2022-12-06 at 1 44 11 PM
Screen Shot 2022-12-06 at 1 43 57 PM

Been super busy the past several days but will be free to work on this together any time this week after tonight!

io/fasta/fasta.go
```diff
@@ -66,14 +70,178 @@ type Fasta struct {
...
+	const translationStopChar = '*'
+	for {
+		line, err = p.rd.ReadSlice('\n')
+		totalRead += int64(len(line))
+		p.line++
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				if len(line) > 1 && line[0] != ';' {
+					// EOF-ended fasta. We append line if not empty
+					stopTranslate := bytes.IndexByte(line, translationStopChar)
```
TimothyStiles, line 171: The stop codon only denotes that a gene will stop being transcribed. In practice a single sequence can have many stop codons. Is it safe to remove this block @soypat?

io/fasta/fasta_test.go
```diff
@@ -248,3 +304,86 @@ func TestWrite_error(t *testing.T) {
...
 }
+
+func TestParser(t *testing.T) {
+	parser := NewParser(nil, 256)
+	for testIndex, test := range []struct {
+		content  string
+		expected []Fasta
+	}{
+		{
+			content:  ">humen\nGATTACA\nCATGAT", // EOF-ended Fasta not valid
```
TimothyStiles, line 315: we should probably be more permissive and along EOF-ended Fasta as valid?

TimothyStiles: @soypat my latest changes have this entire suite of tests failing on off-by-one errors where there's always one more fasta than what is expected. Any idea what's going on?

soypat: I'll take a deep dive look today!

io/fasta/fasta.go
soypat: Oh snap, right will do. Just pushed a incorrect solution in that case

io/fasta/fasta_test.go
soypat: I'm hesitant to say EOF ended fastas are OK. This means you can't use io.LimitReader to read fastas at all using poly API since its use will return incomplete fastas without warning. The reason this is important to me is because io.LimitReader is a commonly used foundation for writing software resilient to adversarial user input.

Though I'd also like users who want to parse EOF ended fastas to be able to use poly. This is possible since we can return a valid fasta alongside an io.EOF error when the fasta is immediately terminated by an EOF. This will allow users who need this functionality to build their own parsers who treat the EOF case specially.

Note: ParseNext now (correctly?) parses `data/empty.fasta`, returning a single fasta.

soypat: Maybe we can add a boolean flag to the ParseN, ParseAll, ParseBytes: `acceptEOFTerminatedFastas`?

io/fasta/fasta.go
```diff
@@ -66,14 +70,186 @@ type Fasta struct {
...
+				// Buffer size too small to read fasta line.
+				return Fasta{}, totalRead, fmt.Errorf("line %d too large for buffer, use larger maxLineSize: %w", p.line+1, err)
+			} else if !isEOF {
+				return Fasta{}, totalRead, err // Unexpected error.
+			}
+
+			// So got to this point the line is probably OK, we will return a Fasta.
+			// We may or may not exit with error depending on whether we find a terminator.
+			sequence = append(sequence, line...)
+			fastaIsTerminated := bytes.IndexByte(line, '*') >= 0
```
soypat, line 191: So just checking, although base.fasta uses a stop codon at the end of the fasta sequence, it could easily have more genes that follow?

TimothyStiles: Yes. The Stop codon is denoting the stopped biological transcription/translation of a sequence not the computational representation of the sequence.

io/fasta/fasta.go
```diff
@@ -66,14 +70,181 @@ type Fasta struct {
...
+			isEOF := errors.Is(err, io.EOF)
+			if isSkippable {
+				if isEOF {
+					// got EOF on a empty or commented line.
+					err = nil
+				}
+				break
+			} else if errors.Is(err, bufio.ErrBufferFull) {
+				// Buffer size too small to read fasta line.
+				return Fasta{}, totalRead, fmt.Errorf("line %d too large for buffer, use larger maxLineSize: %w", p.line+1, err)
```
TimothyStiles, line 183: How would we make this fail? I've tried adjusting the parser struct's max line size variable to 1 but still am not able to get this to be thrown.

io/fasta/fasta.go
```diff
@@ -66,14 +70,181 @@ type Fasta struct {
...
+			continue
+		}
+		// If we got to this point we are currently inside of the fasta
+		// sequence contents. We append line to what we found of sequence so far.
+		sequence = append(sequence, line...)
+	} // parse loop ends here.
+
+	// Parsing ended. Check for inconsistencies.
+	if lookingForName {
+		return Fasta{}, totalRead, fmt.Errorf("did not find fasta start '>', got to line %d: %w", p.line, err)
```
TimothyStiles, line 225: I can't seem to reach this code while parsing with an empty fasta file. Is there a way to do this?

io/fasta/fasta.go
```diff
@@ -66,14 +70,181 @@ type Fasta struct {
...
+	}
+	if !lookingForName && len(sequence) == 0 {
```
TimothyStiles, line 227: Can't reach with test using an empty fasta file.

io/fasta/fasta_test.go
TimothyStiles: > Maybe we can add a boolean flag to the ParseN, ParseAll, ParseBytes: acceptEOFTerminatedFastas?

This feels a little clunky but may be the way to do it. Should this variadic or explicit?

TimothyStiles: > Note: ParseNext now (correctly?) parses data/empty.fasta, returning a single fasta.

Just made empty.fasta truly empty. It's meant to test the corner cases I just commented on.

io/fasta/fasta.go
soypat: Done!

io/fasta/fasta.go
soypat: Done!

io/fasta/fasta.go
soypat: bufio.Reader has a minimum permissible line length of 16. Test now does that, though I feel most of this "error" coverage is in vain. This error in particular can be avoided using bufio.Reader.ReadLine instead of ReadSlice. It just makes parsing harder.

soypat: That's not to say the maxLineLength argument is in vain, it still provides a way to let users avoid heap allocations during scan by setting an appropiate value.

soypat:
I removed `empty.fasta` since it can be represented as a const string

io/fasta/fasta_test.go
soypat: > This feels a little clunky but may be the way to do it. Should this variadic or explicit?

Maybe we can have parser options? Something like this https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis? Or maybe that's too much- just a struct with fields?

Issue #278, opened 2022-10-19
rebase can contain multiple references


Koeng101:
https://github.com/TimothyStiles/poly/blob/f2dcf099fea761c9d00d7fadbe50e3d92652e71c/io/rebase/rebase.go#L242-L246

Current parser version only parses a single reference. Enzymes like BsaI have multiple references.

rkrishnasanka:
Yeah even the test is off, I checked the rebase test file, it looks like it has enzymes with multiple references and it still clears it

palmerito0:
As far as I can tell none of the tests actually check the contents of the resulting `enzymeMap` after parsing, just the fact that parsing proceeds without error.

PR #286, opened 2022-12-17
Fixed the issue `rebase can contain multiple references`

rkrishnasanka:
Fixed the issue #278 . 

# Changelog:
- Rewrote the commercial availability section to use a RegEx based parser that makes it more resilient to malformed / mis-indented files
- Updated the basic structure to capture the availability and the dates as pair entries in the main enzyme data structure
- Cleaned up all depreciated `ioutils` references
- Added a parser test with a minimal test cases 

# Reviewer Questions
1. Do I need to add more tests for this PR ?
2. Can I get support to verify that the JSON dump I'm using is correct ?
3. Do we need to add a stress test with the large rebase test case ? How do we validate it ?
4. Is there a way to generate malformed rebase files to test for panics ?

# Recommendations for the future:
- We need to move science-related data and commercial data to separate data structures that are not clubbed together
- We need to standardize parser writing to use something like ANTLR. Logic writing and debugging will be hard to maintain in the future.

### DIFF
--- a/io/fasta/fasta.go
+++ b/io/fasta/fasta.go
@@ -16,7 +16,6 @@ import (
 	"bytes"
 	"compress/gzip"
 	"io"
-	"io/ioutil"
 	"os"
 	"strings"
 )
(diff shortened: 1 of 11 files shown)


rkrishnasanka:
@TimothyStiles you wanna close this one ? Its fairly small

PR #288, opened 2023-02-03
adding new alignment package and global alignment via Needleman-Wunsch

TimothyStiles:
Poly needs a basic alignment algo to compare sequences. [Needleman-Wunsch](https://en.wikipedia.org/wiki/Needleman%E2%80%93Wunsch_algorithm) is about as basic as it gets as far as I know and its implementation is only 70 lines or so.

I'm a little unsure of what sort of corner cases I should use to test it and I'm interested to hear if there's a better way to structure calls and pass parameters to `align.NeedlemanWunsch`. Otherwise everything is clean and code coverage is 100%.

-Tim


TimothyStiles:
Just added the local alignment algorithm SmithWaterman as well but have yet to implement a traceback to return and it just returns the distance metric score by itself.

TimothyStiles:
I'm not completely satisfied but this is *good enough* to ship. Documentation is okay but needs more. I'd like a whole section on alignment algos but they aren't exactly my forte so any guest documentation would be greatly appreciated.

I've moved some SmithWaterman related stuff to two other branches. I somewhat blackboxed getting my implementation to produce the score correctly but am I little lost on how to get the traceback to produce alignments.

Including things like blosum matrices would be a nice update but I don't personally need it now. Would be happy to include if someone makes a PR!

-Tim

Issue #290, opened 2023-02-14
Implement substitution matrices for alignment


TimothyStiles:
Alignment algorithms compare two sequences and produce a distance metric, "score", based on how similar they are.

This score usually accounts for characters that match, characters that don't, and any gaps between matching characters.

In the case of nucleotides matching and not matching can be a simple +1, -1 for anything that matches or mismatches respectively but in the case of protein sequences these weights can (and probably should) vary depending on the chemical similarities between groups of amino acids.

I know of the blosum substitution matrices and know that biogo has a few along with NeedlemanWunsch and SmithWaterman implementations. My thought would be to cite and use at least the matrices they provide and maybe implement our own alignment algos that would be easier to maintain longterm.

rkrishnasanka:
Citing and reusing would be the best way to go about it.

TimothyStiles:
link to biogo's matrices that should be cited: https://github.com/biogo/biogo/blob/master/align/matrix

PR #292, opened 2023-02-23
Smith-Waterman Algorithm

ragnorc:


align/align.go
```diff
@@ -91,6 +82,74 @@ func NeedlemanWunsch(stringA string, stringB string, scoring Scoring) (int, stri
...
+	alignB := ""
+	i := maxScoreRow
+	j := maxScoreCol
+	for matrix[i][j] > 0 {
+		if matrix[i][j] == matrix[i-1][j-1]+score(stringA[i-1], stringB[j-1], scoring) {
+			alignA = string(stringA[i-1]) + alignA
+			alignB = string(stringB[j-1]) + alignB
+			i--
+			j--
+		} else if matrix[i][j] == matrix[i-1][j]+scoring.GapPenalty {
```
TimothyStiles, line 128: this condition isn't covered by tests

align/align.go
```diff
@@ -91,6 +82,74 @@ func NeedlemanWunsch(stringA string, stringB string, scoring Scoring) (int, stri
...
+			alignA = string(stringA[i-1]) + alignA
+			alignB = "-" + alignB
+			i--
+		} else if matrix[i][j] == matrix[i][j-1]+scoring.GapPenalty {
+			alignA = "-" + alignA
+			alignB = string(stringB[j-1]) + alignB
+			j--
+		} else {
```
TimothyStiles, line 136: is there any case where this can happen? Shouldn't panic but should return an error if this case is possible.

align/align.go
ragnorc: Good point. Just changed the test case to have gaps on both alignment strings: [98b16bf](https://github.com/TimothyStiles/poly/pull/292/commits/98b16bf216a97262726d7427884661a02c7cdf11)

align/align.go
ragnorc: I am pretty sure this case should never happen. It would mean that the matrix was constructed erroneously or there is a bug somewhere.

TimothyStiles:
@ragnorc this was super helpful! I've added some documentation and formatting and refactored some variables. Going to merge this and then we can work on adding weight matrices mentioned in #290.

ragnorc:
> Thank you @ragnorc this PR was excellent! I've added some documentation and formatting and refactored some variables. Going to merge this and then we can work on adding weight matrices mentioned in #290.

Awesome, thank you! Working on the weight matrices now.

PR #295, opened 2023-03-09
Seqfold

folago:
Add seqfold package, Go port of the python project at: https://github.com/Lattice-Automation/seqfold


seqfold/cmd/seqfold/main.go
```diff
@@ -0,0 +1,74 @@
+package main
```
TimothyStiles, line 1: should probably ship command line tool in separate repo?

seqfold/seqfold.go
```diff
@@ -0,0 +1,95 @@
...
+	NN              BpEnergy
+	TERMINAL_MM     BpEnergy
+	TRI_TETRA_LOOPS BpEnergy
+}
+
+type IJ struct {
+	I, J int
+}
+
+// A single structure with a free energy, description, and inward children
```
TimothyStiles, line 44: think there's a more descriptive name than, "Struct"? Little too similar to struct keyword.

seqfold/fold.go
```diff
@@ -0,0 +1,925 @@
...
+				e_sum += w_cache[i2][j2].E
+				branches = append(branches, tb...)
+			}
+		}
+
+		structs[len(structs)-1].E -= e_sum
+		return append(structs, branches...)
+	}
+
+	// this is unreachable also in the python code
```
TimothyStiles, line 911: probably should just delete this?

seqfold/go.mod
```diff
@@ -0,0 +1,14 @@
+module gitlab.com/folago/seqfold.go
```
TimothyStiles, line 1: should probably be removed in favor of project directory go.mod?

seqfold/go.sum
```diff
@@ -0,0 +1,19 @@
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
TimothyStiles, line 1: probably should be removed in favor of project directory go.sum

seqfold/utils.go
```diff
@@ -0,0 +1,91 @@
...
+// RoundFloat runds at the given decimal place
+func RoundFloat(number float64, decimalPlace int) float64 {
+	// Calculate the 10 to the power of decimal place
+	temp := math.Pow(10, float64(decimalPlace))
+	// Multiply floating-point number with 10**decimalPlace and round it
+	// Divide the rounded number with 10**decimalPlace to get decimal place rounding
+	return math.Round(number*temp) / temp
+}
+
+func isDNA(seq string) bool {
```
TimothyStiles, line 47: may want to make this a part of the `checks` package

seqfold/utils.go
```diff
@@ -0,0 +1,91 @@
...
+		case 'A', 'C', 'T', 'G':
+			continue
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+func isRNA(seq string) bool {
```
TimothyStiles, line 59: may want to make this a part of the `checks` package

seqfold/.gitignore
```diff
@@ -0,0 +1 @@
+cmd/seqfold/seqfold
```
TimothyStiles, line 1: this should probably be removed.

TimothyStiles (changes requested):
This is awesome @folago. I've made comments as to what could be removed, or integrated into another package. Going to look into adding more biological context into the comments, renaming confusing variables, writing examples tests, and overall unit test coverage next.

seqfold/seqfold.go
folago: Agreed, it tried to keep the code as close to the python original but this is really confusing.

`NucleicAcidStructure` would be the most appropriate name IMHO even if it's a bit long.
I'll make this change but I'm open to suggestion in case you were thinking on a better name.

seqfold/utils.go
folago: Sounds a bit overkill for just two functions but 🤷🏻

seqfold/utils.go
folago: as above

folago:
Done.

What about README and LICENCE? Keep? Remove? Change?

folago:
Ah also I made fixups, I usually push more fixups until it's OK an then rebase and force push the branch.
I am not sure how you prefer things here, also I can split in more commits if you like.
Let me know.

folago:
Just realized what you meant by the checks 😂

seqfold/seqfold.go
rkrishnasanka: I think long names are okay as long as they tell the story right

TimothyStiles:
@folago go linter is picking up some redundant code and other things.

https://github.com/TimothyStiles/poly/actions/runs/4441863344/jobs/7863721360?pr=295#step:3:71

seqfold/dna.go
```diff
@@ -0,0 +1,519 @@
+package seqfold
+
+var DNA_COMPLEMENT = Comp{'A': 'T', 'T': 'A', 'G': 'C', 'C': 'G', 'N': 'N'}
```
TimothyStiles, line 3: Can use the complement tools in `transform` for this.

seqfold/dna.go
```diff
@@ -0,0 +1,519 @@
...
+// Inferred from:
+// Supplemental Material: Annu.Rev.Biophs.Biomol.Struct.33:415-40
+// doi: 10.1146/annurev.biophys.32.110601.141800
+// The Thermodynamics of DNA Structural Motifs
+// SantaLucia and Hicks, 2004
+var DNA_MULTIBRANCH = MultiBranchEnergies{A: 2.6, B: 0.2, C: 0.2, D: 2.0}
+
+// The Thermodynamics of DNA Structural Motifs
+// SantaLucia and Hicks, 2004
+var DNA_NN = BpEnergy{
```
TimothyStiles, line 15: what does NN stand for?

TimothyStiles: I should put together an explainer for the energy parameters.

seqfold/dna.go
```diff
@@ -0,0 +1,519 @@
...
+	".T/GA": {-4.1, -13.1},
+	".A/TT": {2.9, 10.4},
+	".C/TG": {-4.4, -13.1},
+	".G/TC": {-5.2, -15.0},
+	".T/TA": {-3.8, -12.6},
+}
+
+// the energies are the same for each loop stack in the
+// reverse complementary direction
+func init() {
```
TimothyStiles, line 185: what is this initializing?

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+//     temp: The temperature the Fold takes place in, in Celsius
+// Returns:
+//     List[Struct]: A list of structures. Stacks, bulges, hairpins, etc.
+func Fold(seq string, temp float64) []NucleicAcidStructure {
+	//v_cache, w_cache := cache(seq, temp)
+	//n := len(seq)
+	fc, err := NewFoldingContext(seq, temp)
+	if err != nil {
+		fmt.Printf("%+v\n", err)
+		os.Exit(1)
```
TimothyStiles, line 33: should return error instead of `os.Exit(1)`

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+	for i, wrow := range wCache {
+		for j, s := range wrow {
+			cache[i][j] = s.E
+		}
+	}
+
+	return cache, nil
+}
+
+// Get the dot bracket notation for a secondary structure.
```
TimothyStiles, line 88: need to find explainer of dot bracket notation.

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+// Args:
+//     seq: The sequence being folded
+//     i: The start index
+//     j: The end index (inclusive)
+//     temp: The temperature in Kelvin
+//     v_cache: Free energy cache for if i and j bp
+//     w_cache: Free energy cache for lowest energy structure from i to j. 0 otherwise
+// Returns:
+//     float: The free energy for the subsequence from i to j
+func W(i, j int, fc FoldContext) NucleicAcidStructure {
```
TimothyStiles, line 132: I forget what `W` stands for and we should probably rename it to something more readable. Same goes fo `V`

TimothyStiles: these single letter variables from the original paper + seqfold implementation are funkayy. Should find a way to rename them.

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+// Args:
+//     seq: The sequence being folded
+//     i: The start index
+//     j: The end index (inclusive)
+//     temp: The temperature in Kelvin
+//     v_cache: Free energy cache for if i and j bp. INF otherwise
+//     w_cache: Free energy cache for lowest energy structure from i to j. 0 otherwise
+//     emap: Energy map for DNA/RNA
+// Returns:
+//    float: The minimum energy folding structure possible between i and j on seq
```
TimothyStiles, line 173: ^^ these are useful for renaming args to something more human readable.

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+// InternalLoop calculates the free energy of an internal loop.
+//
+// The first and last bp of both left and right sequences
+// are not themselves parts of the loop, but are the terminal
+// bp on either side of it. They are needed for when there's
+// a single internal looping bp (where just the mismatching
+// free energies are used)
+// Note that both left and right sequences are in 5' to 3' direction
+// This is adapted from the "Internal Loops" section of SantaLucia/Hicks, 2004
+// Args:
```
TimothyStiles, line 519: rename variables and args based on this mapping for InternalLoop

seqfold/fold.go
```diff
@@ -0,0 +1,875 @@
...
+//     query_len: Length of element without known free energy value
+//     known_len: Length of element with known free energy value (d_g_x)
+//     d_g_x: The free energy of the element known_len
+//     temp: Temperature in Kelvin
+// Returns:
+//     float: The free energy for a structure of length query_len
+func JS(query_len, known_len int, d_g_x, temp float64) float64 {
+	gas_constant := 1.9872e-3
+	return d_g_x + 2.44*gas_constant*temp*math.Log(float64(query_len)/float64(known_len))
```
TimothyStiles, line 754: expand this to be more commented and readable

seqfold/fold_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+		fc, err = NewFoldingContext(seq, 37)
+		require.NoError(t, err)
+
+		struc = W(i, j, fc)
+		assert.InDelta(t, struc.E, -4.2, 0.2)
+	})
+}
+
+func ExampleFold() {
```
TimothyStiles, line 189: create new example_test.go file for ExampleFold()

seqfold/utils.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func abs(x int) int {
+	if x < 0 {
+		return -x
+	}
+	return x
+}
+
+func Reverse(s string) string {
```
TimothyStiles, line 30: Know it's silly but there should be a reverse function in `transform` that you can call via `transform.Reverse(<string>)`

seqfold/seqfold.go
```diff
@@ -0,0 +1,157 @@
...
+	HAIRPIN_LOOPS   LoopEnergy
+	MULTIBRANCH     MultiBranchEnergies
+	INTERNAL_LOOPS  LoopEnergy
+	INTERNAL_MM     BpEnergy
+	NN              BpEnergy
+	TERMINAL_MM     BpEnergy
+	TRI_TETRA_LOOPS BpEnergy
+}
+
+type Interval struct {
```
TimothyStiles, line 43: what does this interval type do? What are I and J? Start and end of subsequence?

seqfold/seqfold.go
```diff
@@ -0,0 +1,157 @@
...
+	seq = strings.ToUpper(seq)
+
+	// figure out whether it's DNA or RNA, choose energy map
+	var emap Energies
+	switch {
+	case checks.IsDNA(seq):
+		emap = DNA_ENERGIES
+	case checks.IsRNA(seq):
+		emap = RNA_ENERGIES
+	default:
```
TimothyStiles, line 125: Tired so maybe this logic is already solved but what happens when a sequence contains neither T or U?

seqfold/seqfold.go
TimothyStiles: @folago I'd say it's "done" and ready to be refactored to be idiomatic!

seqfold/dna.go
folago: Yes, but it seems to work on only for DNA, what about RNA?

Also is `unsafe` really necessary? But that's another issue altogether.

seqfold/dna.go
folago: I have no idea what `NN` stands for 😅

seqfold/dna.go
folago: It is adding the reverse of each sequence in the the energy maps, since they have the same energy IIUC.
Also I left the python code commented out as reference to be sure I was doing the right thing.

seqfold/fold.go
folago: Ah yes, I was still thinking in terms of command line tools as the python code.

There are also a bunch of panics I left in place of python `assert`.
I am not entirely sure how to deal with those, it is probably an error we cannot recover from so either we bubble it up to the caller or recover at some point.

I would prefer the error all the way but it's probably a big refactor.

seqfold/fold.go
folago: [This](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4041456/#:~:text=The%20most%20common%20encoding%20for,code%20for%20an%20unpaired%20base%20'.) might help but I did not manage to find a good source that explains the notation.

seqfold/seqfold.go
folago: It is considered DNA since it is the first to be checked.
Not much else we can do in this case IMHO...

seqfold/fold.go
folago: I did a pass, I'll do another later this week when I should have some more time.

seqfold/seqfold.go
folago: Yes, AFAIU it is basically a subsequence that holds an inner/child structure, I'm open to suggestion for the naming.
Something like:

```go
type Subsequence struct{
    Start, End int
}
```

seqfold/fold.go
folago: Renamed it with a better name, I don't really know what this is doing 😅

seqfold/dna.go
folago: Unrolled the maps and removed the `init()`

folago:
I'd say it's time for a rebase? And maybe I can split the PR in more commits with a decent commit message.

The only missing part that I can see (a part from some more docs and comments) is the complements since `transform.ComplementBase` only seems to complement DNA

folago:
Sorry for the wild rebases but it was getting a too messy for my taste, I have a local copy of the last version in case you want to keep a track of things (especially your commit @TimothyStiles).

Some changes in no particular order:
- no more panics, just passing errors as it is supposed to be
- small docs improvement
- fix linter false positive

This https://github.com/TimothyStiles/poly/pull/295#issuecomment-1490725652 is IMHO still open, should I just use `transform.ComplementBase` for the DNA? Should I add `transform.ComplementBaseRNA`?

TimothyStiles:
> Sorry for the wild rebases but it was getting a too messy for my taste, I have a local copy of the last version in case you want to keep a track of things (especially your commit @TimothyStiles).
> 
> Some changes in no particular order:
> 
>     * no more panics, just passing errors as it is supposed to be
> 
>     * small docs improvement
> 
>     * fix linter false positive
> 
> 
> This [#295 (comment)](https://github.com/TimothyStiles/poly/pull/295#issuecomment-1490725652) is IMHO still open, should I just use `transform.ComplementBase` for the DNA? Should I add `transform.ComplementBaseRNA`?

No problem on the rebases. Using transform.ComplementBase for DNA and adding `transform.ComplementBaseRNA` would be excellent @folago!

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+// Figure 2B in Zuker and Stiegler, 1981
+// Args:
+//
+//		seq: The sequence being folded
+//		i: The start index
+//		j: The end index (inclusive)
+//	 fc: The FoldingContext for this sequence
+//
+// Returns the free energy for the subsequence from i to j
+func W(i, j int, fc FoldContext) (NucleicAcidStructure, error) {
```
TimothyStiles, line 103: should this be unexported?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+// If i and j don't bp, store and return INF.
+// See: Figure 2B of Zuker, 1981
+// Args:
+//
+//		i: The start index
+//		j: The end index (inclusive)
+//	 fc: The FoldingContext for this sequence
+//
+// Returns the minimum energy folding structure possible between i and j on seq
+func V(i, j int, fc FoldContext) (NucleicAcidStructure, error) {
```
TimothyStiles, line 155: should this be unexported?

fold/utils.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func max[T constraints.Ordered](a, b T) T {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func min[T constraints.Ordered](a, b T) T {
```
TimothyStiles, line 16: not tested or used?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+	fc, err := NewFoldingContext(seq, temp)
+	if err != nil {
+		return nil, fmt.Errorf("error creating folding context: %w", err)
+	}
+
+	// get the minimum free energy structure out of the cache
+	return Traceback(0, len(seq)-1, fc), nil
+}
+
+// Fold the sequence and return just the delta G of the structure
```
TimothyStiles, line 38: does `Energy` just return the minimum free energy while `Fold` returns MFE and structure?

fold/dna.go
```diff
@@ -0,0 +1,607 @@
...
+	"sym":      {H: 0, S: -1.4},
+	"tini":     {H: 0.2, S: -5.7}}
+
+// Internal mismatch table (DNA)
+// Allawi & SantaLucia (1997), Biochemistry 36: 10581-10594
+// Allawi & SantaLucia (1998), Biochemistry 37: 9435-9444
+// Allawi & SantaLucia (1998), Biochemistry 37: 2170-2179 *
+// Allawi & SantaLucia (1998), Nucl Acids Res 26: 2694-2701 *
+// Peyret et al. (1999), Biochemistry 38: 3468-3477 *
+var DNA_INTERNAL_MM = BPEnergy{"AA/AT": Energy{H: 4.7, S: 12.9},
```
TimothyStiles, line 40: trying to change fold.FoldEnergy to fold.Energy but everytime I do I get a compiler error here:

````
initialization cycle for DNA_INTERNAL_MM
	./dna.go:40:5: DNA_INTERNAL_MM refers to
	./fold.go:45:6: Energy refers to
	./fold.go:28:6: Fold refers to
	./seqfold.go:128:6: NewFoldingContext refers to
	./dna.go:596:5: DNA_ENERGIES refers to
	./dna.go:40:5: DNA_INTERNAL_MMgo
  ```

File: fold/dna.go [hunk 27]
```diff
@@ -0,0 +1,607 @@
+package fold
+
+var DNA_COMPLEMENT = map[byte]byte{'A': 'T', 'T': 'A', 'G': 'C', 'C': 'G', 'N': 'N'}
+
+var DNA_MULTIBRANCH = MultiBranchEnergies{A: 2.6, B: 0.2, C: 0.2, D: 2.0}
+
+// The Thermodynamics of DNA Structural Motifs
+// SantaLucia and Hicks, 2004
+var DNA_NN = BPEnergy{"AA/TT": Energy{H: -7.6, S: -21.3},
```
TimothyStiles, line 9: what does NN mean in DNA_NN?

fold/dna.go
```diff
@@ -0,0 +1,607 @@
...
+	27: {0, -19.7},
+	28: {0, -20.0},
+	29: {0, -20.0},
+	30: {0, -20.3},
+}
+
+var DNA_ENERGIES = Energies{
+	BULGE_LOOPS:     DNA_BULGE_LOOPS,
+	COMPLEMENT:      DNA_COMPLEMENT,
+	DE:              DNA_DE,
```
TimothyStiles, line 599: what is DE?

fold/dna.go
```diff
@@ -0,0 +1,607 @@
...
+	HAIRPIN_LOOPS:   DNA_HAIRPIN_LOOPS,
+	MULTIBRANCH:     DNA_MULTIBRANCH,
+	INTERNAL_LOOPS:  DNA_INTERNAL_LOOPS,
+	INTERNAL_MM:     DNA_INTERNAL_MM,
+	NN:              DNA_NN,
```
TimothyStiles, line 604: what is NN?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+func MinimumFreeEnergy(seq string, temp float64) (float64, error) {
+	NucleicAcidStructures, err := Fold(seq, temp)
+	if err != nil {
+		return 0, fmt.Errorf("error folding: %w", err)
+	}
+	summedEnergy := 0.0
+	for _, structure := range NucleicAcidStructures {
+		summedEnergy += structure.E
+	}
+	return RoundFloat(summedEnergy, 2), nil
```
TimothyStiles, line 54: why the round float here?

fold/seqfold.go
```diff
@@ -0,0 +1,173 @@
...
+	i, j := "", ""
+	if len(structure.Inner) > 0 {
+		i, j = fmt.Sprint(structure.Inner[0].Start), fmt.Sprint(structure.Inner[0].End)
+	}
+	return fmt.Sprintf("%4s %4s % 6.2f  %-15s", i, j, structure.Energy, structure.Desc)
+}
+
+// MultiString returns all the fields as strings, this is useful to output the
+// result in a tabular fashion using the same format string.
+func (structure NucleicAcidStructure) MultiString() (string, string, string, string) {
```
TimothyStiles, line 95: tests don't reach. Dead code?

fold/seqfold.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+	return structure.Energy == other.Energy
+}
+
+func (structure NucleicAcidStructure) Valid() bool {
+	return structure.Energy != math.Inf(1) && structure.Energy != math.Inf(-1)
+}
+
+func (structure NucleicAcidStructure) String() string {
```
TimothyStiles, line 85: tests don't reach. Dead code?

fold/seqfold.go
```diff
@@ -0,0 +1,173 @@
...
+	InternalLoops LoopEnergy
+	INTERNAL_MM   BPEnergy
+	NN            BPEnergy
+	TERMINAL_MM   BPEnergy
+	TriTetraLoops BPEnergy
+}
+
+// Subsequence represent an interval of bases in the sequence that can contain
+// a inward structure.
+type Subsequence struct {
```
TimothyStiles, line 54: should this be exported? probably should be internal struct.

fold/seqfold.go
```diff
@@ -0,0 +1,173 @@
...
+	Start, End int
+}
+
+// A single structure with a free energy, description, and inward children
+type NucleicAcidStructure struct {
```
TimothyStiles, line 59: should probably be internal struct.

fold/seqfold.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+// MultibranchEnergies holds the a, b, c, d in a linear multi-branch energy
+// change function.
+// Inferred from:
+// Supplemental Material: Annu.Rev.Biophs.Biomol.Struct.33:415-40
+// doi: 10.1146/annurev.biophys.32.110601.141800
+// The Thermodynamics of DNA Structural Motifs, SantaLucia and Hicks, 2004
+type MultibranchEnergies struct {
+	A, B, C, D float64
```
TimothyStiles, line 18: what does each letter individually represent?

TimothyStiles: looking at the section cited and there's actually a fifth parameter not used?

* A - number of helices in the loop
* B - number of unpaired nucleotides in the loop
* C - coxial stacking in the loop
* D - terminal mismatch contributions
* E - base composition of the unpaired nucleotides (probably neglible?)

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+		e_Multibranch = a + d
+	}
+
+	// energy of min-energy neighbors
+	e := e_Multibranch + e_sum
+
+	// pointer to next structures
+	if helix {
+		// branches.pop()
+		branches = branches[:len(branches)-1]
```
TimothyStiles, line 520: get the feeling this should be optimized?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+	// heuristic for speeding this up
+	// from https://www.ncbi.nlm.nih.gov/pubmed/10329189
+	isolated_outer := true
+	if start > 0 && end < len(foldContext.Seq)-1 {
+		isolated_outer = foldContext.Energies.Complement[foldContext.Seq[start-1]] != foldContext.Seq[end+1]
+	}
+	isolated_inner := foldContext.Energies.Complement[foldContext.Seq[start+1]] != foldContext.Seq[end-1]
+
+	if isolated_outer && isolated_inner {
+		foldContext.V[start][end] = NucleicAcidStructure{Energy: 1600}
```
TimothyStiles, line 175: why this magic nummber of `1600` for energy?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+		return mm_left + mm_right, nil
+	}
+	var d_h, d_s, dG float64
+	// apply a penalty based on loop size
+	if en, ok := foldContext.Energies.InternalLoops[loop_len]; ok {
+		d_h, d_s = en.EnthalpyH, en.EntropyS
+		dG = DeltaG(d_h, d_s, foldContext.T)
+	} else {
+		// it's too large an internal loop, extrapolate
+		en := foldContext.Energies.InternalLoops[30]
```
TimothyStiles, line 570: why this magic number `30`? Should probably be internal package constant.

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+		d_h, d_s = en.EnthalpyH, en.EntropyS
+		dG = DeltaG(d_h, d_s, foldContext.T)
+		dG = JacobsonStockmayer(loop_len, 30, dG, foldContext.T)
```
TimothyStiles, line 573: why this magic number `30`? Should probably be internal package constant.

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+	e1 := NucleicAcidStructure{Energy: hp, Description: "HAIRPIN:" + p}
+	if end-start == 4 { // small hairpin; 4bp
+		foldContext.V[start][end] = e1
+		foldContext.W[start][end] = e1
+		return foldContext.V[start][end], nil
+	}
+
+	n := len(foldContext.Seq)
+	e2 := NucleicAcidStructure{Energy: math.Inf(1)}
+	for i1 := start + 1; i1 < end-4; i1++ {
```
TimothyStiles, line 193: what are these magic numbers for?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+//
+//	NucleicAcidStructures: A list of NucleicAcidStructure, usually from the fold function
+//
+// # Returns the dot-bracket notation of the secondary structure
+//
+// Dot-bracket notation, consisting in a balanced parentheses string composed
+// by a three-character alphabet {.,(,)}, that can be unambiguously converted
+// in the RNA secondary structure. See example_test.go for a small example.
+func DotBracket(NucleicAcidStructures []NucleicAcidStructure) string {
+	maxj := 0
```
TimothyStiles, line 68: what is maxj?

fold/fold.go
```diff
@@ -0,0 +1,895 @@
...
+//		end: The end index (inclusive)
+//	 foldContext: The FoldContext for this sequence
+//
+// Returns the free energy for the subsequence from start to end
+func W(start, end int, foldContext FoldContext) (NucleicAcidStructure, error) {
+	if !foldContext.W[start][end].Equal(DefaultStructure) {
+		return foldContext.W[start][end], nil
+	}
+
+	if end-start < 4 {
```
TimothyStiles, line 108: why less than four. What is this magic number?

TimothyStiles: ah. Looks like the original 1981 Zuker paper relies on "pentanucleotides". Off by one?

fold/fold.go
folago: Depends on the use for this package and its usefulness outside the folding. 

We could start not exporting anything that is not needed in a `cmd` and then export in case it's needed. This way we keep the API surface small and less prone to breaking changes?

fold/utils.go
folago: Leftovers, removing...

fold/fold.go
folago: It was there in the python code, I assume it was rounded to be pretty printed.
I think we can safely remove it.

fold/seqfold.go
folago: Handy method to replicate the same output as the python code, I can add a test but I'm not entirely sure if this code belongs here or not.

fold/fold.go
folago: Need to read the papers to answer that, but yes it should be in a constant.

fold/fold.go
folago: Seems the be the max of the `End` of the internal subsequences.

fold/seqfold.go
TimothyStiles: should probably just remove it and `string`

folago:
> No problem on the rebases. Using transform.ComplementBase for DNA and adding `transform.ComplementBaseRNA` would be excellent @folago!

I was looking at the transform package again and found a couple thing I think are worth mentioning:
- `ReverseComplement` and `Complement` have different behavior than `ComplementBase` since the former two do not use the latter but access the complement table directly. So a missing value in the first two function results in a `0` while `ComplementBase` gives a space ` `.
- Test should be, IMHO, repeatable and predicable random test are neither.
- Shouldn't we just use a `strings.Builder` instead of punching a hole trough the abstraction offered by the standard library and avoid `unsafe`?
- Since you added the align package I'd add the gap `-` to the table.

Apologies for being pedantic, I am up for submitting changes to what I mentioned just wanted to have the green light.

I also can just copy/paste the code as is to have a reverse for RNA and we can talk about this another time (or never if this is good enough™)

fold/fold.go
```diff
@@ -0,0 +1,923 @@
...
+	if !foldContext.unpairedMinimumFreeEnergyW[start][end].Equal(defaultStructure) {
+		return foldContext.unpairedMinimumFreeEnergyW[start][end], nil
+	}
+
+	if end-start < 4 {
+		foldContext.unpairedMinimumFreeEnergyW[start][end] = invalidStructure
+		return foldContext.unpairedMinimumFreeEnergyW[start][end], nil
+	}
+
+	w1, err := unpairedMinimumFreeEnergyW(start+1, end, foldContext)
```
TimothyStiles, line 141: what are these different tiers of unpaired minimum free energies? what can we call them besides w1, w2, w3?

fold/fold.go
folago: I'm not sure what you mean

fold/fold.go
folago: Apparently it's a heuristic from https://www.ncbi.nlm.nih.gov/pubmed/10329189

Moving to a const.

fold/fold.go
folago: It is the max length we have for precomputed energies, above that we use JacobsonStockmayer.
Moving to a const.

fold/fold.go
folago: Not sure we are `0` based

folago: No, it makes not sense the `0` based I think you are right and it should be `<= 4`

fold/fold.go
folago: yes found some better names from the paper

fold/fold.go
folago: I think it's again the limit of pentanucleotide or less cannot create a structure

folago: There are two more magic `4` that I'm not sure about

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+		}
+
+	}
+
+	if unpaired < 0 {
+		return defaultStructure, fmt.Errorf("multibranch: subsequence (%d, %d, %d): unpaired < 0", start, end, k)
+	}
+
+	// penalty for unmatched bp and multi-branch
+	a, b, c, d := foldContext.energies.multibranch.helicesCount, foldContext.energies.multibranch.unpairedCount, foldContext.energies.multibranch.coaxialStackCount, foldContext.energies.multibranch.terminalMismatchCount
```
TimothyStiles, line 491: should rename single letter variables.

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+		// of the helix as though it was a dangling end
+		// if there's only one bp, it goes to whichever
+		// helix (this or the next) has the more favorable energy
+		unpairedLeft := 0
+		unpairedRight := 0
+		danglingEnergy := 0.0
+		if index == len(branches)-1 && !helix {
+			// pass
+		} else if ij3 == subsequence {
+			unpairedLeft = i2 - j1 - 1
```
TimothyStiles, line 434: rename the `i2` et al

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+		// it's too large an internal loop, extrapolate
+		foldEnergy := foldContext.energies.internalLoops[maxLenPreCalulated]
+		enthalpyHDifference, entropySDifference = foldEnergy.enthalpyH, foldEnergy.entropyS
+		dG = deltaG(enthalpyHDifference, entropySDifference, foldContext.temp)
+		dG = jacobsonStockmayer(loopLength, maxLenPreCalulated, dG, foldContext.temp)
+	}
+
+	// apply an asymmetry penalty
+	loopAsymmetry := math.Abs(float64(loopLeftIndex - loopRightIndex))
+	dG += 0.3 * loopAsymmetry
```
TimothyStiles, line 562: what is 0.3?

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+
+	pairedMismatchRightEnergy := pair(foldContext.seq, i1-1, i1, j1+1, j1)
+	foldEnergy = foldContext.energies.terminalMismatches[pairedMismatchRightEnergy]
+	enthalpyHDifference, entropySDifference = foldEnergy.enthalpyH, foldEnergy.entropyS
+	dG += deltaG(enthalpyHDifference, entropySDifference, foldContext.temp)
+
+	return dG, nil
+}
+
+// stack return the free energy of a stack
```
TimothyStiles, line 578: returns*

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+	//    return 0.0
+	for _, x := range []int{start, i1, end, j1} {
+		if x >= len(foldContext.seq) {
+			return 0
+		}
+	}
+
+	paired := pair(foldContext.seq, start, i1, end, j1)
+	// if any(x == -1 for x in [start, i1, end, j1]):
+	for _, x := range []int{start, i1, end, j1} {
```
TimothyStiles, line 606: rename `i1` et al

TimothyStiles: rename x

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+//	 foldContext: The FoldingContext for this sequence
+//
+// Returns the free energy increment from the hairpin structure
+func hairpin(start, end int, foldContext context) (float64, error) {
+	if end-start < minLenForStruct {
+		return math.Inf(1), nil
+	}
+
+	hairpinSeq := foldContext.seq[start : end+1]
+	hairpinLength := len(hairpinSeq) - 2
```
TimothyStiles, line 684: why -2?

fold/fold.go
```diff
@@ -0,0 +1,879 @@
...
+	// add penalty for a terminal mismatch
+	energy, ok := foldContext.energies.terminalMismatches[paired]
+	if hairpinLength > 3 && ok {
+		enthalpyHDifference, entropySDifference := energy.enthalpyH, energy.entropyS
+		dG += deltaG(enthalpyHDifference, entropySDifference, foldContext.temp)
+	}
+
+	// add penalty if length 3 and AT closing, formula 8 from SantaLucia, 2004
+	if hairpinLength == 3 && (hairpinSeq[0] == 'A' || hairpinSeq[len(hairpinSeq)-1] == 'A') {
+		dG += 0.5 // convert to entropy
```
TimothyStiles, line 722: should we make these constants?

TimothyStiles (commented):
quick review of style related stuff. More to add.

fold/fold.go
folago: Or even from Santaluia 2004:
image

fold/fold.go
folago: The closing BP? But I din't find a good explanation.

folago:
I pushed few more style changes.
I have a question on the packages layout, if eventually linearfold is merged will the `fold` package undergo a breaking change?

TimothyStiles:
> I pushed few more style changes. I have a question on the packages layout, if eventually linearfold is merged will the `fold` package undergo a breaking change?

that's actually something I've been wondering too. Thoughts were.

Rename the `Fold` function to `Zuker` so that it would be called `fold.Zuker` and we could name the linear fold function as `fold.Linear`.

Think that would work @folago?

transform/transform.go
```diff
@@ -88,7 +88,6 @@ var complementTable = [256]byte{
 	'R': 'Y',
 	'S': 'S',
 	'T': 'A',
-	'U': 'A',
```
TimothyStiles, line 91: reason for removing this?

TimothyStiles:
@folago MERGED.

Could use a little more test coverage but this is great and I'm trying to get all outstanding PRs merged before I move poly to a new org.

jjti:
This is sick, @folago. I didn't review in detail, but I love it. Would be curious about the speed up of this vs the python version

folago:
> This is sick, @folago. I didn't review in detail, but I love it. Would be curious about the speed up of this vs the python version

Thanks! 
I did not do a thorough comparison, just a quick test using cython to run seqfold and the speed seemed comparable. 

I did not look at any profiles though, so maybe there are some easy wins.

PR #296, opened 2023-03-20
fix comment typos

ysung6:
typo fixes for comments


seqhash/seqhash.go
```diff
@@ -70,8 +69,8 @@ import (
 type SequenceType string
 
 const (
-	DNA SequenceType = "DNA"
-	RNA SequenceType = "RNA"
+	DNA     SequenceType = "DNA"
```
TimothyStiles, line 72: looks like there's extra space being inserted here?

PR #298, opened 2023-03-23
Slow5

Koeng101:
This PR will add a slow5 format parser, from https://github.com/hasindu2008/slow5tools

Doesn't have a blow5 parser, but slow5 is good enough for my use case - @TimothyStiles could I get a review?


Koeng101:
I should probably add a writer as well

io/slow5/slow5.go
```diff
@@ -0,0 +1,289 @@
...
+				for headerNum := 1; headerNum < len(values); headerNum++ {
+					headerMap[headerNum] = values[headerNum]
+				}
+				start = true
+				continue
+			}
+		}
+
+		// Once we have the read headers, start to parse the actual reads
+		if start {
```
matiasinsaurralde, line 187: I think this line can be removed so that all the block that comes after it is unindented and will be always called when `!start`, then you add a `continue` after line 183 just to ensure the loop continues, so:

```go
if !start {
  ...
  continue
}

// Once we have the read headers, start to parse the actual reads
var newRead Read
for valueIndex := 0; valueIndex < len(values); valueIndex++ {
...
```

io/slow5/slow5.go
```diff
@@ -0,0 +1,289 @@
...
+			var newRead Read
+			for valueIndex := 0; valueIndex < len(values); valueIndex++ {
+				fieldValue := headerMap[valueIndex]
+				if values[valueIndex] == "." {
+					continue
+				}
+				switch fieldValue {
+				case "read_id":
```
matiasinsaurralde, line 195: Thoughts about making `read_id`, `read_group`, etc. constants and add them to the top of the file?

io/slow5/slow5.go
```diff
@@ -0,0 +1,289 @@
...
+				if err != nil {
+					return []Header{}, err
+				}
+				numReadGroups = uint32(numReadGroupsUint)
+				for id := uint32(0); id < numReadGroups; id++ {
+					emptyMap := make(map[string]string)
+					headers = append(headers, Header{Slow5Version: slow5Version, ReadGroupId: id, Attributes: emptyMap})
+				}
+			}
+		} else {
```
matiasinsaurralde, line 119: We can add continue after line 118 and unindent the `else` block too:
```go
if numReadGroups == 0 {
  ...
  continue
}
// The block that occurs when numReadGroups isn't zero continues here
```

io/slow5/slow5.go
```diff
@@ -0,0 +1,289 @@
...
+		values := strings.Split(line, "\t")
+		if len(values) < 2 {
+			return []Header{}, fmt.Errorf("Got following line without tabs: %s", line)
+		}
+
+		// First, we need to identify the number of read groups. This number will be the length of our
+		// ReadGroups output, and we will need it for iteration through the rest of the header.
+		if numReadGroups == 0 {
+			switch values[0] {
+			case "#slow5_version":
```
matiasinsaurralde, line 106: Same suggestion as the other, these could be constants 👍

io/slow5/slow5.go
Koeng101: Hmmm... does that make the code more readable? The constants would essentially just be the string, without quotes around it. When reading the code, you'd have to then refer to the const if you want to know what it is actually checking for, but the const is essentially just that string.

Koeng101:
96.4% test coverage (the only things I don't test is the write function, because it is difficult to test that.

Last rewrite was quite large - I made the parser similar to the fasta parser method developed by @soypat in #284 . I agree, it is a lot clearer!

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+		failureReasons[failureReason] = failureIndex
+	}
+
+	// Iterate over reads. This is reading from a channel, and will end
+	// when the channel is closed.
+	for read := range reads {
+		// converts []int16 to string
+		rawSignalString := strings.Trim(strings.Replace(fmt.Sprint(read.RawSignal), " ", ",", -1), "[]")
+		// Look at above output.Write("#read_id ... for the values here.
+		_, err = output.Write([]byte(fmt.Sprintf("%s\t%d\t%g\t%g\t%g\t%g\t%d\t%s\t%d\t%d\t%d\t%g\t%d\t%s\n", read.ReadId, read.ReadGroupId, read.Digitisation, read.Offset, read.Range, read.SamplingRate, read.LenRawSignal, rawSignalString, read.StartTime, read.ReadNumber, read.StartMux, read.MedianBefore, failureReasons[read.EndReason], read.ChannelNumber)))
```
soypat, line 403: Suggest using fmt.Fprintf instead of the `Write` method- might save you quite a bit of allocations depending on how well the compiler optimized the current line i.e.:
```go
_, err = fmt.Fprintf(output, "%s\t%d\t%g\t%g\t%g\t%g\t%d\t%s\t%d\t%d\t%d\t%g\t%d\t%s\n", read.ReadId, //...
```

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+// Write writes a list of headers and a channel of reads to an output.
+func Write(headers []Header, reads <-chan Read, output io.Writer) error {
+	// First, write the slow5 version number
+	slow5Version := headers[0].Slow5Version
+	_, err := output.Write([]byte(fmt.Sprintf("#slow5_version\t%s\n", slow5Version)))
+	if err != nil {
+		return err
+	}
+	// Then, write the number of read groups (ie, the number of headers)
+	_, err = output.Write([]byte(fmt.Sprintf("#num_read_groups\t%d\n", len(headers))))
```
soypat, line 332: Same as below, use fmt.Fprintf

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+
+Keoni
+
+******************************************************************************/
+
+// Write writes a list of headers and a channel of reads to an output.
+func Write(headers []Header, reads <-chan Read, output io.Writer) error {
+	// First, write the slow5 version number
+	slow5Version := headers[0].Slow5Version
+	_, err := output.Write([]byte(fmt.Sprintf("#slow5_version\t%s\n", slow5Version)))
```
soypat, line 327: Ditto below

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = output.Write([]byte("#read_id	read_group	digitisation	offset	range	sampling_rate	len_raw_signal	raw_signal	start_time	read_number	start_mux	median_before	end_reason	channel_number\n"))
+	if err != nil {
+		return err
+	}
+	// Parse failure reasons
+	failureReasonsString := "unknown,partial,mux_change,unblock_mux_change,data_service_unblock_mux_change,signal_positive,signal_negative"
+	failureReasons := make(map[string]int)
```
soypat, line 392: failuresReasons is a constant map on every run? Maybe this should be a global variable to avoid runtime overhead?

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+			case "#slow5_version":
+				slow5Version = values[1]
+			case "#num_read_groups":
+				numReadGroupsUint, err := strconv.ParseUint(values[1], 10, 32)
+				if err != nil {
+					return parser, []Header{}, err
+				}
+				numReadGroups = uint32(numReadGroupsUint)
+				for id := uint32(0); id < numReadGroups; id++ {
+					emptyMap := make(map[string]string)
```
soypat, line 145: You could inline emptyMap's creation with the append below. that is write `... Attributes: make(map[string]string)}`

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+		case "start_mux":
+			startMux, err := strconv.ParseUint(values[valueIndex], 10, 8)
+			if err != nil {
+				newRead.Error = fmt.Errorf("Failed to convert start_mux '%s' to uint on line %d. Got error: %s", values[valueIndex], parser.line, err)
+			}
+			newRead.StartMux = uint8(startMux)
+		case "median_before":
+			medianBefore, err := strconv.ParseFloat(values[valueIndex], 64)
+			if err != nil {
+				newRead.Error = fmt.Errorf("Failed to convert median_before '%s' to float on line %d. Got error: %s", values[valueIndex], parser.line, err)
```
soypat, line 281: Might be worth replacing the final `%s` formatter verb with `%w` for wrapping functionality- very small change and you give users extra error handling functionality.

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+			attributeRow.Write([]byte(attributeValue))
+		}
+		headerAttributeStrings = append(headerAttributeStrings, attributeRow.String())
+	}
+	// Sort the header strings
+	sort.Strings(headerAttributeStrings)
+
+	// Write the header attribute strings to the output
+	for _, headerAttributeString := range headerAttributeStrings {
+		_, err = output.Write([]byte(fmt.Sprintf("%s\n", headerAttributeString)))
```
soypat, line 374: ditto below. fmt.Fprintf

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Write the read headers
+	// These are according to the slow5 specifications
+	_, err = output.Write([]byte("#char*	uint32_t	double	double	double	double	uint64_t	int16_t*	uint64_t	int32_t	uint8_t	double	enum{unknown,partial,mux_change,unblock_mux_change,data_service_unblock_mux_change,signal_positive,signal_negative}	char*\n"))
```
soypat, line 382: Might be a good idea to also apply `fmt.Fprintln` here. The reason being is that I'm not sure if a heap allocation is done for the string- I've asked on the performance channel in the gophers slack to check...

io/slow5/slow5.go
```diff
@@ -0,0 +1,409 @@
...
+	failureReasons := make(map[string]int)
+	for failureIndex, failureReason := range strings.Split(failureReasonsString, ",") {
+		failureReasons[failureReason] = failureIndex
+	}
+
+	// Iterate over reads. This is reading from a channel, and will end
+	// when the channel is closed.
+	for read := range reads {
+		// converts []int16 to string
+		rawSignalString := strings.Trim(strings.Replace(fmt.Sprint(read.RawSignal), " ", ",", -1), "[]")
```
soypat, line 401: why is read.RawSignal inside of a `fmt.Sprint` call? Couldn't you just remove the `fmt.Sprint` call?

soypat (changes requested):
Couple of performance related changes suggested and miscellaneous

io/slow5/slow5.go
Koeng101: Because `RawSignal    []int16`. The sprintf makes it into a string, and then I just do some replacements to convert to the kind of list expected in slow5.

Alternatively, I could do a string builder or something.

Koeng101: I don't really know if it improves performance, but I modified this to use a stringBuilder

io/slow5/slow5.go
```diff
@@ -0,0 +1,440 @@
...
+	"signal_positive":                 true,
+	"signal_negative":                 true,
+}
+
+// Parser is a flexible parser that provides ample
+// control over reading slow5 sequences.
+// It is initialized with NewParser.
+type Parser struct {
+	// reader keeps state of current reader.
+	reader       bufio.Reader
```
TimothyStiles, line 102: should these capitalized and exported?

Koeng101: Idk, should they? I don't have a strong opinion either way.

soypat: I think safest path is to leave them unexported, this protects the underlying implementation allowing it to change if we find a better way of parsing. Like Gophers say: You can always export them in the future if it comes to that. But unexporting them is a breaking change.

io/slow5/slow5_test.go
```diff
@@ -0,0 +1,223 @@
...
+			break
+		}
+		outputReads = append(outputReads, read)
+	}
+	if outputReads[0].RawSignal[0] != 430 {
+		t.Errorf("Expected first outputRead to have a raw_signal of 430. Got: %d", outputReads[0].RawSignal[0])
+	}
+}
+
+func ExampleNewParser() {
```
TimothyStiles, line 46: add all example tests to their own file and slow5_test package.

io/slow5/data/example.slow5
TimothyStiles, line 1: where is this coming from? <- real data from what a nanopore sequencing reaction <- @Koeng101 has more details.

PR #299, opened 2023-03-23
seqhash2

Koeng101:
This PR is initializing seqhash_v2, something I've wanted to do for a while. 

Basically, all seqhash_v2 adds is the ability to have base modifications (important for some newer cloning methods I'm using) and non-homologous doublestranded DNA (important for cloning parts from small annealed oligos).

DNA / RNA is NOT the same if it has modifications! Dcm/dam modifications matter for doing restriction digestions! tRNAs need modifications to function as expected!


Koeng101:
Eh, not doing right now.

PR #307, opened 2023-05-17
Circular restriction cut

Koeng101:
Very simple PR. I noticed CutWithRestrictionEnzyme didn't work across sequence origins, so I fixed that.


Koeng101:
Crap, accidentally messed up merge... one moment

PR #312, opened 2023-05-29
updated example optimize

### LINKED ISSUE #306: JCVI codon table generation is flat-out-wrong
It returns:
```
* has the following weights:
[  "TAA: 4680",  "TAG: 1369"]
```
This is impossible, as JCVI has <500 genes.

Koeng101:
Fixes #306 

Makes optimization example clearer


Koeng101:
This example also still assumes that start codons are a part of the sequence. I need to change that

PR #316, opened 2023-06-07
Improve fragment interface

Koeng101:
Depends on #309 

I added existingOverhangs as an input to Fragment. As it turns out, this is a fairly common operation, and I think it is clearest if it is added to the API.


TimothyStiles:
@Koeng101 MERGED.

Koeng101:
IMG_1274

PR #322, opened 2023-08-02
Ensure Non-Met start codons map to Met on translation

### LINKED ISSUE #305: Translate should have a setting to set first codon to M
testTranslation, _ := codon.Translate(gfpDnaSequence, codon.GetCodonTable(11))

This fails in the case of  `gfpDnaSequence := "TTGGCT...TAA"`, which is perfectly valid for a start codon.

carreter:
Also makes Translate() return an error if an invalid start codon is received

Fixes #305.


synthesis/codon/codon_test.go
```diff
@@ -21,6 +21,26 @@ func TestTranslation(t *testing.T) {
...
+}
+
+func TestTranslationErrorsOnIncorrectStartCodon(t *testing.T) {
+	badSequence := "GGG"
+
+	if _, gotErr := Translate(badSequence, GetCodonTable(11)); gotErr == nil {
+		t.Errorf("Translation should return an error if given an incorrect start codon")
+	}
+}
+
```
Koeng101, line 43: Great tests!

Koeng101 (approved):
This looks great to me, and is exactly the kind of checks we needed. Thanks!

PR #328, opened 2023-08-15
Add svb compression to slow5

Koeng101:
This PR adds a simple way to use StreamVByte compression with slow5. This is the primary way that slow5 files are compressed, so if you are using poly to io between different file stores or pipe between different programs, StreamVByte greatly helps - I've seen reduction in total database size (storing fastq and slow5 reads) from 12Gb to 7.1Gb using just this one weird trick (database engineers hate him)

The code changes have 100% code coverage, and I'm pretty sure I wrote more comments than actual code (not including tests).


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/google/go-cmp v0.5.8
+	github.com/koeng101/svb v0.0.0-20230815034912-d6737f9ed8b8
```
TimothyStiles, line 7: this dependency has assembly in it?

Koeng101: Yes. It make it go zoom!

Koeng101:
@TimothyStiles Any thoughts?

io/slow5/slow5.go
```diff
@@ -196,7 +198,9 @@ func NewParser(r io.Reader, maxLineSize int) (*Parser, []Header, error) {
 	return parser, headers, nil
 }
 
-// ParseNext parses the next read from a parser.
+// ParseNext parses the next read from a parser. Note: you should check
+// Read.Error for errors that happen within a read, while the overall error
+// for errors that happen in the parser.
```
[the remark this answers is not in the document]
carreter: Actually, I don't think I agree with this behavior regardless. Why do we require clients to check two separate places for errors? It would make sense to put errors in the Read.Error field if we were parsing multiple reads at a time and wanted to not fail if there is an individual bad read, but we are reading one at a time here. 

IMO it would be best to also return read errors in the `error` return value to ensure the client notices. Thoughts?

io/slow5/slow5.go
```diff
@@ -437,3 +441,50 @@ func Write(headers []Header, reads <-chan Read, output io.Writer) error {
...
+
+Stream VByte paper: https://doi.org/10.48550/arXiv.1709.08990
+
+Keoni
+
+******************************************************************************/
+
+// SvbCompressRawSignal takes a read and converts its raw signal field to two
+// arrays: a mask array and a data array. Both are needed for decompression.
+func SvbCompressRawSignal(rawSignal []int16) (mask, data []byte) {
```
carreter, line 471: Can we have these functions be methods of Read/another type? As it stands it's not clear to me how a client is supposed to take advantage of them.

Maybe use struct inheritance to define a new CompressedRead class and have a method that goes from CompressedRead -> Read and vice versa.

io/slow5/slow5_test.go
```diff
@@ -204,3 +205,31 @@ func TestWrite(t *testing.T) {
...
+
+func TestSvb(t *testing.T) {
+	file, _ := os.Open("data/example.slow5")
+	defer file.Close()
+	const maxLineSize = 2 * 32 * 1024
+	parser, _, _ := NewParser(file, maxLineSize)
+	var outputReads []Read
+	for {
+		read, err := parser.ParseNext()
+		if err != nil {
```
carreter, line 217: Please make this an explicit check for `io.EOF`. As written, this test will pass (even though it shouldn't!) if there are any I/O errors while reading from disk.

io/slow5/slow5.go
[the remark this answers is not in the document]
carreter: @TimothyStiles as well for visibility

io/slow5/slow5.go
Koeng101: I'm using these functions for taking the rawSignal in and out of an SQL database, so having it be a method doesn't work as well as a raw function. 

`CompressedRead` I think would just be a blow5 read rather than slow5 read, which we SHOULD implement at some point. (there is a whole binary format specifically for doing those compressed reads well, with some real performance improvements)

io/slow5/slow5_test.go
```diff
@@ -204,3 +205,31 @@ func TestWrite(t *testing.T) {
 		t.Errorf("Example and test write are different")
 	}
 }
+
+func TestSvb(t *testing.T) {
+	file, _ := os.Open("data/example.slow5")
+	defer file.Close()
+	const maxLineSize = 2 * 32 * 1024
+	parser, _, _ := NewParser(file, maxLineSize)
```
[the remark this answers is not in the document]
Koeng101: Fixed in ioToBio

io/slow5/slow5_test.go
Koeng101: I think we already have that check in `TestParseReads`, but I'll add it here too

PR #331, opened 2023-08-17
parser wasnt iterating lines, so error messages were nonsensical. Fixed

Koeng101:
Stupid simple addition. parser.line wasn't iterating in slow5 parser, so error messages would be a bit non-nonsensical. Fixed this with 1 line.


Koeng101:
hmmm... too lazy to figure out the right git commands, just gonna open another pr

PR #338, opened 2023-08-24
improvement of fastq and fasta

Koeng101:


Koeng101:
Don'tcha hate when tests fail from files that aren't a part of your commit

Koeng101:
Moving to #339

Issue #327, opened 2023-08-08
slow5 doesn't detect newline


Koeng101:
If there is a newline at the end of a slow5 file, the parser detects this as an additional line / read.

TimothyStiles:
@Koeng101 was this ever fixed?

Issue #325, opened 2023-08-07
fastq de-aligned


Koeng101:
https://github.com/TimothyStiles/poly/blob/f587e72731bdfebe2e452604707a4efe8906672a/io/fastq/fastq.go#L120C38-L120C38

Still figuring it out, but noting here: fastq becomes unaligned while reading. Not sure. This causes the entire thing to get messed up.

Koeng101:
Found it - https://github.com/TimothyStiles/poly/blob/f587e72731bdfebe2e452604707a4efe8906672a/io/fastq/fastq.go#L210

*(*string)(unsafe.Pointer(&sequence)), // Stdlib strings.Builder.String() does this so it *should* be safe.

This is NOT safe.

Koeng101:
Alright, I got this error again. 

If you redefine the sequence in terms of a new sequence, it is just fine.

```go
sequence = line[:len(line)-1] // Exclude newline delimiter.
    var newSequence []byte
    for _, char := range sequence {
        newSequence = append(newSequence, char)
        if !strings.Contains("ATGCN", string(char)) {
            return Fastq{}, totalRead, errors.New("Only letters ATGCN are allowed for DNA/RNA in fastq file. Got letter: " + string(char))
        }
    }
```

I have no idea why, but this becomes a thing production. It takes a long ass time to find.

PR #339, opened 2023-08-24
io->bio

Koeng101:
This is a fairly large PR to standardize our parsing and writing of files, using generics to implement higher level functions on simplified interfaces that are standardized across all our readers/writers: genbank, gff, fasta, fastq, slow5, sam, uniprot, rebase, pileup


Koeng101:
This is a work-in-progress, btw

carreter:
Why the `io` to `bio` rename?

Koeng101:
> Why the `io` to `bio` rename?

So that the name doesn't conflict with standard library io

bio/fasta/example_test.go
```diff
@@ -50,7 +54,7 @@ func ExampleBuild() {
 // ExampleWrite shows basic usage of the  writer.
 func ExampleWrite() {
 	fastas, _ := fasta.Read("data/base.fasta")       // get example data
-	_ = fasta.Write(fastas, "data/test.fasta")       // write it out again
+	_ = fasta.WriteFile(fastas, "data/test.fasta")   // write it out again
```
TimothyStiles, line 57: should be just `write`

bio/bio.go
```diff
@@ -0,0 +1,23 @@
...
+Package bio provides utilities for reading and writing sequence data.
+*/
+package bio
+
+import (
+	"io"
+)
+
+/*
+This package is supposed to be empty and only exists to provide a doc string.
```
TimothyStiles, line 11: no longer true

bio/bio.go
```diff
@@ -0,0 +1,23 @@
...
+Otherwise its namespace would collide with Go's native IO package.
+*/
+
+type Parser[T any, TH any] interface {
```
TimothyStiles, line 15: What are `T` and `TH` here? Some more explanation on the need for an interface and how it works would be very helpful.

bio/example_test.go
```diff
@@ -22,7 +22,7 @@ func Example() {
 	// Poly can also output these file formats. Every file format has a corresponding Write function.
 	_ = gff.Write(gffInput, "test.gff")
 	_ = genbank.Write(gbkInput, "test.gbk")
-	_ = fasta.Write(fastaInput, "test.fasta")
+	_ = fasta.WriteFile(fastaInput, "test.fasta")
```
TimothyStiles, line 25: why `WriteFile` instead of `Write`?

bio/fasta/fasta.go
```diff
@@ -56,12 +55,6 @@ Keoni
 
 ******************************************************************************/
 
-var (
```
TimothyStiles, line 59: these were used for testing.

bio/bio.go
Koeng101: Types, as are used in generics. I've added in `DataTypes` and `DataTypeHeaders` to be a little more descriptive. While `T` is fairly idiomatic go, `TH` is kinda pushing it, so I think longer descriptions are nice.

bio/fasta/fasta.go
Koeng101: All of these were used in tests that are moving out of `fasta` and into `bio`, where those tests can test all packages at once using generic functions.

bio/example_test.go
Koeng101: `Write` in Golang is typically not used on paths (as far as I've seen), but on `io.Writer` interfaces. This change to `WriteFile` when using paths makes it more idiomatic. 

`Write` generally is also being moved into the bio package where we can more effectively test across all parser packages.

bio/fastq/fastq.go
```diff
@@ -177,6 +170,13 @@ func (parser *Parser) ParseNext() (Fastq, int64, error) {
 		return Fastq{}, totalRead, fmt.Errorf("empty fastq sequence for %q,  got to line %d: %w", seqIdentifier, parser.line, err)
 	}
 	sequence = line[:len(line)-1] // Exclude newline delimiter.
+	var newSequence []byte
+	for _, char := range sequence {
+		newSequence = append(newSequence, char)
+		if !strings.Contains("ATGCN", string(char)) {
```
carreter, line 176:
```suggestion
		if !strings.ContainsRune("ATGCN", rune(char)) {
```

bio/bio.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (p *Parser[DataType, DataTypeHeader]) Header() (DataTypeHeader, int64, error) {
+	return p.LowLevelParser.Header()
+}
+
+func (p *Parser[DataType, DataTypeHeader]) ParseN(countN int) ([]DataType, error) {
+	return nil, nil
+}
+
+func (p *Parser[DataType, DataTypeHeader]) Parse() ([]DataType, error) {
```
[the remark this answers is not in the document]
Koeng101: `ParseAll` includes the header. For a lot of formats, the header doesn't really matter (fasta, fastq, etc), for some it matters a lot (slow5, etc)

bio/bio.go
```diff
@@ -0,0 +1,176 @@
...
+Lower level interfaces
+
+******************************************************************************/
+
+type LowLevelParser[DataType fasta.Fasta, DataTypeHeader fasta.Header] interface {
+	Header() (DataTypeHeader, int64, error)
+	Next() (DataType, int64, error)
+}
+
+type Writer interface {
```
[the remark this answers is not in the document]
Koeng101: Ty, didn't look there, changed the type signatures.

bio/bio.go
```diff
@@ -0,0 +1,176 @@
...
+	return data, header, nil
+}
+
+func (p *Parser[DataType, DataTypeHeader]) ParseConcurrent(channel chan<- DataType) error {
+	return nil
+}
+
+/******************************************************************************
+
+Writer functions
```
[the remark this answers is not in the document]
Koeng101: `io.MultiWriter()` has a type signature of `MultiWriter(writers ...Writer) Writer`, whereas `WriteAll` has a type signature of `WriteAll(data []io.WriterTo, header io.WriterTo, w io.Writer) error`. The writerTo are different, and generally they do different things: `WriteAll` writes many records + header to a single file, while `io.MultiWriter` splits one write to many writers, kinda like `tee`.

bio/bio.go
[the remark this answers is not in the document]
carreter: Ah, gotcha. Maybe call it `ParseWithHeader` instead?

bio/bio.go
```diff
@@ -4,20 +4,173 @@ Package bio provides utilities for reading and writing sequence data.
...
+)
+
+// DefaultMaxLineLength variables are defined for performance reasons. While
+// parsing, reading byte-by-byte takes far, far longer than reading many bytes
+// into a buffer. In golang, this buffer in bufio is usually 64kb. However,
+// many files, especially slow5 files, can contain single lines that are much
+// longer. We use the default maxLineLength from bufio unless there is a
+// particular reason to use a different number.
+const defaultMaxLineLength int = bufio.MaxScanTokenSize // 64kB is a magic number often used by the Go stdlib for parsing.
+var (
```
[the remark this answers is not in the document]
Koeng101: Implemented, good catch.

bio/bio.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (p *Parser[DataType, DataTypeHeader]) Next() (DataType, int64, error) {
+	return p.LowLevelParser.Next()
+}
+
+func (p *Parser[DataType, DataTypeHeader]) Header() (DataTypeHeader, int64, error) {
+	return p.LowLevelParser.Header()
+}
+
+func (p *Parser[DataType, DataTypeHeader]) ParseN(countN int) ([]DataType, error) {
```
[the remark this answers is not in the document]
Koeng101: Implemented all. Mostly it was just copy pasting the copy-pastes around from other parsers...

bio/bio.go
```diff
@@ -4,20 +4,173 @@ Package bio provides utilities for reading and writing sequence data.
...
-	Next() (T, error)
-	MaxLineCount() int64
+/******************************************************************************
+Aug 30, 2023
+
+Lower level interfaces
+
+******************************************************************************/
+
+type LowLevelParser[DataType fasta.Fasta, DataTypeHeader fasta.Header] interface {
```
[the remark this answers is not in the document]
Koeng101: Hmm I agree we should probably change `LowLevelParser`, but Concrete isn't quite there for me. My first reaction was "Concrete, like a foundation, like the lowest level?"

bio/bio.go
```diff
@@ -0,0 +1,176 @@
...
+	return NewParser(format, file)
+}
+
+/******************************************************************************
+
+Parser higher-level functions
+
+******************************************************************************/
+
+func (p *Parser[DataType, DataTypeHeader]) Next() (DataType, int64, error) {
```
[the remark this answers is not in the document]
Koeng101: I had the same thoughts actually, I'm like 50/50 on `int64`. The reason I added that was to stay consistent with `io.Read`, and for the fact that you could use the numbers as a seek for larger files (like save the offsets)

Koeng101:
`fasta` parser testing is now at 98.5%, other than the Scanner non-EOF error, which I can't really figure out how to test after a very brief trying-to-figure-out. I removed a ton of code, and everything should be fairly simple now.

bio/bio.go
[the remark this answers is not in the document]
Koeng101: `WriteFile()` is just legacy - pretty much all our previous packages have it. @TimothyStiles thoughts on including a `WriteFile()`? I'm leaning towards removing it, since os.Open is pretty simple.

bio/bio.go
```diff
@@ -0,0 +1,195 @@
...
+
+type Parser[DataType fasta.Record, DataTypeHeader fasta.Header] struct {
+	LowLevelParser LowLevelParser[DataType, DataTypeHeader]
+}
+
+var emptyParser Parser[fasta.Record, fasta.Header] = Parser[fasta.Record, fasta.Header]{}
+
+func NewParser(format Format, r io.Reader) (*Parser[fasta.Record, fasta.Header], error) {
+	var maxLineLength int
+	switch format {
```
carreter, line 73: Since we have a map of max lengths now, this switch isn't necessary.

bio/bio.go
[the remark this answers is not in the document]
carreter: We are already not consistent with `io.Read`. The `int64` is only really necessary to know how many bytes were read into the buffer you pass to `io.Read`.

bio/bio.go
```diff
@@ -0,0 +1,195 @@
...
+		return []DataType{}, DataTypeHeader{}, err
+	}
+	data, err := p.Parse()
+	if err != nil {
+		return []DataType{}, DataTypeHeader{}, err
+	}
+	return data, header, nil
+}
+
+func (p *Parser[DataType, DataTypeHeader]) ParseConcurrent(channel chan<- DataType) error {
```
carreter, line 155: As written, this non-current and blocking. You probably want to wrap the for loop in a `go func(){ ... }()` to make it nonblocking and call it `ParseToChannel()` instead. Make sure to note whether or not it is block in that case.

bio/fasta/fasta.go
```diff
@@ -51,84 +50,47 @@ Keoni
...
-//}
-
 // Parser is a flexible parser that provides ample
 // control over reading fasta-formatted sequences.
 // It is initialized with NewParser.
 type Parser struct {
 	// reader keeps state of current reader.
-	reader bufio.Reader
-	line   uint
+	reader     bufio.Scanner
```
carreter, line 67: Call this `scanner` instead

bio/bio.go
Koeng101: Concurrency in term of parsing / IO is kind of a lie though, because you are literally reading the file with ` io.Reader`. I do agree though, bad name, I like the name `ParseToChannel`

(removed the `close(channel)` for below stuff)

The change to `go func(){}()` would make it nonblocking, but also make error checking more difficult. I'm thinking of using this with something like [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup), so that you can begin parsing multiple files while still having error handling in a pretty simple-to-understand way.

In the case that you are parsing one file, IO will likely be your limiting factor, and a blocking function is pretty easy to use and understand how it works. As you need to scale, you can just wrap the overall function in `go func(){}()`.


I am open to making it non-blocking tho - is there any particular reason you have in mind for making it non-blocking?

bio/bio.go
Koeng101: True, fixing!

bio/bio.go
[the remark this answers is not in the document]
Koeng101: Makes sense, changing

bio/bio.go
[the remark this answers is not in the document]
Koeng101: > The int64 is only really necessary to know how many bytes were read into the buffer you pass to io.Read

True, I'm convinced. Changing.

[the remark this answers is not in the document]
Koeng101: removed

bio/fastq/fastq.go
```diff
@@ -0,0 +1,212 @@
...
+// Parser is a flexible parser that provides ample
+// control over reading fastq-formatted sequences.
+// It is initialized with NewParser.
+type Parser struct {
+	// reader keeps state of current reader.
+	reader bufio.Reader
+	line   uint
+}
+
+// Header returns a identifier with nothing and nil.
```
Koeng101, line 60: remove 'a identifier'

Koeng101:
fixes #325

Koeng101:
fixes #327

Koeng101:
I'm gonna say this is probably no longer a work-in-progress. @TimothyStiles review?

bio/fasta/fasta_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+const (
+	// This fasta stream contains no Fasta.
+	emptyFasta = "testing\natagtagtagtagtagatgatgatgatgagatg\n\n\n\n\n\n\n\n\n\n\n"
+)
+
+func TestHeader(t *testing.T) {
+	_, err := NewParser(nil, 256).Header()
+	if err != nil {
+		t.Errorf("Header should never be nil")
```
carreter, line 19:
```suggestion
		t.Errorf("Unexpected error while parsing header: %v", err)
```

carreter (commented):
Can we move all the `bio/<format>/data` files to `test/data` or `testdata/`?

bio/bio.go
```diff
@@ -0,0 +1,307 @@
...
+// have headers. Next() returns a record/read/line from the file format, and
+// terminates on an io.EOF error.
+//
+// Next() may terminate with an io.EOF error with a nil DataType or with a
+// full DataType, depending on where the EOF is in the actual file. A check
+// for this is needed at the last Next(), when it returns an io.EOF error. A
+// pointer is used to represent the difference between a null DataType and an
+// empty DataType.
+type LowLevelParser[DataType fasta.Record | fastq.Read | slow5.Read | pileup.Line, DataTypeHeader fasta.Header | fastq.Header | slow5.Header | pileup.Header] interface {
+	Header() (*DataTypeHeader, error)
```
carreter, line 70: Why does `LowLevelParser` return pointers and `Parser` return values?

bio/bio.go
```diff
@@ -0,0 +1,307 @@
...
+	Next() (*DataType, error)
+}
+
+// The following checks that all DataType and DataTypeHeaders implement the io.WriteTo interface.
```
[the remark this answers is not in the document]
Koeng101: Can you link that comment?

bio/bio.go
Koeng101: I think both return pointers. Could you link where `Parser` returns values?

bio/bio.go
```diff
@@ -0,0 +1,307 @@
...
+// fastq, and pileup do not contain headers, while others like sam and slow5 do
+// have headers. Next() returns a record/read/line from the file format, and
+// terminates on an io.EOF error.
+//
+// Next() may terminate with an io.EOF error with a nil DataType or with a
+// full DataType, depending on where the EOF is in the actual file. A check
+// for this is needed at the last Next(), when it returns an io.EOF error. A
+// pointer is used to represent the difference between a null DataType and an
+// empty DataType.
+type LowLevelParser[DataType fasta.Record | fastq.Read | slow5.Read | pileup.Line, DataTypeHeader fasta.Header | fastq.Header | slow5.Header | pileup.Header] interface {
```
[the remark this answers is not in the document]
Koeng101: I renamed to `parserInterface`, because it is the interface that the Parsers have to fulfill

Koeng101:
> Can we move all the `bio/<format>/data` files to `test/data` or `testdata/`?

The difficulty there would be if any of the specific parsers require the testdata, they can't really import `..`. So we'd have to have 2 copies, or we just have different test data for the specific parsers and generic parsers.

bio/bio.go
carreter: Oh my bad I was misreading. Marking resolved.

Koeng101:
> > Can we move all the `bio/<format>/data` files to `test/data` or `testdata/`?
> 
> The difficulty there would be if any of the specific parsers require the testdata, they can't really import `..`. So we'd have to have 2 copies, or we just have different test data for the specific parsers and generic parsers.

I'm completely wrong here actually, I was thinking of go:embed. Yes we can do this

Koeng101:
```
Auto-merging bio/genbank/example_test.go
Auto-merging bio/genbank/genbank_test.go
Auto-merging bio/gff/example_test.go
Auto-merging bio/pileup/pileup.go
CONFLICT (content): Merge conflict in bio/pileup/pileup.go
Auto-merging bio/pileup/pileup_test.go
CONFLICT (content): Merge conflict in bio/pileup/pileup_test.go
Auto-merging bio/polyjson/example_test.go
Auto-merging bio/slow5/slow5.go
Auto-merging bio/uniprot/example_test.go
CONFLICT (content): Merge conflict in bio/uniprot/example_test.go
CONFLICT (modify/delete): io/example_test.go deleted in HEAD and modified in main.  Version main of io/example_test.go left in tree.
CONFLICT (modify/delete): io/fasta/fasta.go deleted in HEAD and modified in main.  Version main of io/fasta/fasta.go left in tree.
Auto-merging seqhash/seqhash_test.go
Auto-merging synthesis/codon/codon_test.go
Auto-merging synthesis/codon/example_test.go
Automatic merge failed; fix conflicts and then commit the result.
```

Good lord, going through and fixing.

bio/example_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	gffInput, _ := gff.Read("../data/ecoli-mg1655-short.gff")
+	gbkInput, _ := genbank.Read("../data/puc19.gbk")
+	//fastaInput, _ := fasta.Read("fasta/data/base.fasta")
+	jsonInput, _ := polyjson.Read("../data/cat.json")
+
+	// Poly can also output these file formats. Every file format has a corresponding Write function.
+	_ = gff.Write(gffInput, "test.gff")
+	_ = genbank.Write(gbkInput, "test.gbk")
+	//_ = fasta.WriteFile(fastaInput, "test.fasta")
```
TimothyStiles, line 31: why is fasta commented out?

bio/example_test.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
+// This is where the integration tests that make effed up cyclic dependencies go.
+func Example() {
+	// Poly can take in basic gff, gbk, fasta, and JSON.
+	// We call the json package "pson" (poly JSON) to prevent namespace collision with Go's standard json package.
+
+	gffInput, _ := gff.Read("../data/ecoli-mg1655-short.gff")
+	gbkInput, _ := genbank.Read("../data/puc19.gbk")
+	//fastaInput, _ := fasta.Read("fasta/data/base.fasta")
```
TimothyStiles, line 25: why is fasta commented out?

bio/example_test.go
```diff
@@ -0,0 +1,234 @@
...
+	parser, _ := bio.NewFastaParser(file)
+
+	records, _ := parser.Parse()
+
+	fmt.Println(records[1].Sequence)
+	// Output: ADQLTEEQIAEFKEAFSLFDKDGDGTITTKELGTVMRSLGQNPTEAELQDMINEVDADGNGTIDFPEFLTMMARKMKDTDSEEEIREAFRVFDKDGNGYISAAELRHVMTNLGEKLTDEEVDEMIREADIDGDGQVNYEEFVQMMTAK*
+}
+
+func Example_readGz() {
+	// ReadGz lets you read gzipped files into a parser.
```
TimothyStiles, line 54: ReadGz isn't a function anymore. Update comment.

bio/example_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	fmt.Println(records[1].Sequence)
+	// Output: ADQLTEEQIAEFKEAFSLFDKDGDGTITTKELGTVMRSLGQNPTEAELQDMINEVDADGNGTIDFPEFLTMMARKMKDTDSEEEIREAFRVFDKDGNGYISAAELRHVMTNLGEKLTDEEVDEMIREADIDGDGQVNYEEFVQMMTAK*
+}
+
+func Example_newParserGz() {
+	// First, lets make a file that is gzip'd, represented by this
+	// buffer.
+	var file bytes.Buffer
+	zw := gzip.NewWriter(&file)
```
TimothyStiles, line 68: `zw` stands for "zip writer". Unclear.

bio/example_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	fileDecompressed, _ := gzip.NewReader(&file) // Decompress the file
+	parser, _ := bio.NewFastaParser(fileDecompressed)
+	records, _ := parser.Parse() // Parse all data records from file
+
+	fmt.Println(records[1].Sequence)
+	// Output: ADQLTEEQIAEFKEAFSLFDKDGDGTITTKELGTVMRSLGQNPTEAELQDMINEVDADGNGTIDFPEFLTMMARKMKDTDSEEEIREAFRVFDKDGNGYISAAELRHVMTNLGEKLTDEEVDEMIREADIDGDGQVNYEEFVQMMTAK*
+}
+
+func ExampleParser_ParseWithHeader() {
```
TimothyStiles, line 90: Am liking the interface overhaul overall 👍

bio/pileup/pileup.go
```diff
@@ -60,12 +56,12 @@ type Pileup struct {
...
-// Parse parses a given Pileup file into an array of Pileup structs.
-func Parse(r io.Reader) ([]Pileup, error) {
-	// 32kB is a magic number often used by the Go stdlib for parsing. We multiply it by two.
-	const maxLineSize = 2 * 32 * 1024
-	parser := NewParser(r, maxLineSize)
-	return parser.ParseAll()
+// Header is a blank struct, needed for compatibility with bio parsers. It contains nothing.
+type Header struct{}
+
+// WriteTo is a blank function, needed for compatibility with bio parsers. It doesn't do anything.
```
TimothyStiles, line 62: Why no writer?

bio/pileup/pileup.go
```diff
@@ -74,47 +70,25 @@ type Parser struct {
 	line   uint
 }
 
+// Header returns a identifier with nothing and nil.
```
TimothyStiles, line 73: ?

bio/bio.go
```diff
@@ -0,0 +1,307 @@
...
+	"bufio"
+	"context"
+	"errors"
+	"io"
+	"math"
+
+	"github.com/TimothyStiles/poly/bio/fasta"
+	"github.com/TimothyStiles/poly/bio/fastq"
+	"github.com/TimothyStiles/poly/bio/pileup"
+	"github.com/TimothyStiles/poly/bio/slow5"
```
TimothyStiles, line 22: No genbank or gff?

TimothyStiles (commented):
Not sure why gff and genbank are not included in `bio.go`.

bio/pileup/pileup.go
Koeng101: There is a writer, `WriteTo` (needed for interface compatibility). Pileup files don't have headers though, so in this case it doesn't do anything.

bio/example_test.go
Koeng101: No `Read` functions anymore.

bio/example_test.go
Koeng101: No `Write` functions anymore (see: `Example_writeAll()`)

bio/example_test.go
Koeng101: Those could be added back in for each and every parser, it's just a bit annoying because how generic initialization works (ie, there would need to be a function for every type, like how parsing is done)

bio/example_test.go
Koeng101: We *could* reimplement them, but the type checking will kinda suck, or we'd have to be super verbose. The reason is that `[]*DataType` does not equal `[]interface{}`, even if `*DataType` DOES equal `interface{}`. Because its a slice, you have to either accept `any` as an input, or you have to do an interface conversion beforehand, which is kinda the worst.

Alternatively, they all implement `io.WriterTo`, so you can just write it to a document using a standard-library  interface.

Koeng101:
@TimothyStiles Just made the genbank parser compatible. All it cost was... nuking all of the tests, because they pretty much all relied on read functions. Those I'll have to add back in later.

Koeng101:
Ah crap, other files depend on `genbank.Read`... I'll fix later.

bio/bio.go
carreter: Actually, why are we returning pointers to the structs instead of the structs themselves?

bio/bio.go
```diff
@@ -0,0 +1,323 @@
...
+// records/reads/lines to a single channel.
+func ManySlow5ToChannel(ctx context.Context, parsers []*Parser[slow5.Read, slow5.Header], channel chan<- slow5.Read) error {
+	return manyToChannel[slow5.Read, slow5.Header](ctx, parsers, channel)
+}
+
+// ManyPileupToChannel takes a list of pileup parsers and pipes all
+// records/reads/lines to a single channel.
+func ManyPileupToChannel(ctx context.Context, parsers []*Parser[pileup.Line, pileup.Header], channel chan<- pileup.Line) error {
+	return manyToChannel[pileup.Line, pileup.Header](ctx, parsers, channel)
+}
```
carreter, line 297: `ParseToChannel` already checks for `ctx.Done()` so this is unnecessary.

bio/bio.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+// ManyFastaToChannel takes a list of FASTA parsers and pipes all
+// records/reads/lines to a single channel.
+func ManyFastaToChannel(ctx context.Context, parsers []*Parser[fasta.Record, fasta.Header], channel chan<- fasta.Record) error {
+	return manyToChannel[fasta.Record, fasta.Header](ctx, parsers, channel)
+}
+
+// ManyFastqToChannel takes a list of FASTQ parsers and pipes all
```
carreter, line 281: This functions are unnecessary as `manyToChannel` is already generic.

carreter (changes requested):
I haven't reviewed the individual formats yet, will come back and do that later.

carreter:
Crap, seems like the line numbers on my last review got messed up because I hadn't synced VSCode with the remote repo. Let me know if you have any questions @Koeng101

bio/bio.go
```diff
@@ -0,0 +1,323 @@
...
+var _ io.WriterTo = (*fastq.Header)(nil)
+var _ io.WriterTo = (*genbank.Header)(nil)
+var _ io.WriterTo = (*slow5.Header)(nil)
+var _ io.WriterTo = (*pileup.Header)(nil)
+
+/******************************************************************************
+
+Higher level parse
+
+******************************************************************************/
```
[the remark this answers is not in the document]
Koeng101: This is a really really good point and I can't believe I didn't think of this, makes a lot of sense. Changing! (gotta go change all the NewParser funcs tho)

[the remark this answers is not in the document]
Koeng101: So the problem is:

The parsers assume that you can return a `Data, io,EOF` and for that data to still be valid. This mainly occurs when an EOF occurs before a newline at the end of some data, which is fairly common in the test cases. In order to use an `io.WriterTo`, we would have to change all the underlying parsers to only return valid `Data` OR `io.EOF`, not AND. 

This is doable, but will take me changing the 5 or so underlying parsers. Overall, it makes sense to me, so I'll do this, but it'll take a little.

bio/bio.go
```diff
@@ -0,0 +1,323 @@
...
+// then optionally closes that channel. If parsing a single file,
+// "keepChannelOpen" should be set to false, which will close the channel once
+// parsing is complete. If many files are being parsed to a single channel,
+// keepChannelOpen should be set to true, so that an external function will
+// close channel once all are done parsing.
+//
+// Context can be used to close the parser in the middle of parsing - for
+// example, if an error is found in another parser elsewhere and all files
+// need to close.
+func (p *Parser[DataType, DataTypeHeader]) ParseToChannel(ctx context.Context, channel chan<- DataType, keepChannelOpen bool) error {
```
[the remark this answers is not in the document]
Koeng101: Well, originally, it was to simplify the use of the channel. Ie, you should be able to run `go ParseToChannel` and then when the parsing is done it'll clean up the channel for you. Otherwise, your for loop wouldn't really end when you start receiving from the channel, and you have no way to know if it terminated, unless you put in your own waitgroup or whatever

bio/bio.go
Koeng101: True. Deleted.

bio/bio.go
Koeng101: You'd have to do a generic initialization on `manyToChannel` though, and I'm not sure we should have our users have to figure out generics

carreter: Is it necessary? I think the compiler might be able to infer the types, at least my IDE was indicating that calling `manyToChannel` didn't require the generic initialization.

Koeng101: Ya know what, I guess I haven't tried, since it could infer the type from the channel input. Lemme test it out, I might be WRONG and DUMB

Koeng101: u right, you don't have to do generic initialization and I am just bad at generics

Koeng101:
- Converted all types to io.Writer in parser. All the parsers have to return is something that fulfills io.Writer for them to be valid. 
- Changed return types on all parsers to NOT ever return an io.EOF and valid `Data` at the same time.
- Genbank merged, all other dependencies fixed.

- [ ] Add back Genbank tests
- [ ] Merge gff
- [ ] Merge uniprot
- [ ] Add comprehensive EOF/newline tests for all parsers.
- [ ] Update comments about `Next()` function

carreter:
Don't forget to credit @folago :

Co-authored-by: name <name@example.com>

carreter:
Fixes #364 .

Koeng101:
Ok, it is more important I think, to get this merged than be perfect. I'm doing the following.

- Removing the gff parser, for now. We can add it back later.
- I have not merged uniprot. We'll have to do that later as we standardize everything
- Add comprehensive EOF/newline tests for all parsers will have to be done later

I will be merging genbank tests and Next comments. Upcoming...

Koeng101:
If we want to go ahead with this, we would spawn 3 new issues:

- Make compatible uniprot parser
- Make compatible gff parser
- Add comprehensive EOF / newline parser tests

abondrn (commented):
Overall looks great, I only had stylistic comments. Due to the size of the PR and it being a pareto improvement (common interface, strictly greater testing, no loss in efficiency) I'm leaning towards approving ASAP. Since all of the key contributors have had their say, for the remaining change requests simply determine whether it is (1) a quick fix (2) issue worthy (3) won't do. No additional features please.

seqhash/seqhash_test.go
```diff
@@ -66,7 +67,10 @@ func TestHash(t *testing.T) {
 }
 
 func TestLeastRotation(t *testing.T) {
-	sequence, _ := genbank.Read("../data/puc19.gbk")
+	file, _ := os.Open("../data/puc19.gbk")
+	defer file.Close()
```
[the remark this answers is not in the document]
Koeng101: I thought that was how it worked, but code I've seen still usually has the defer Close! Do you know why?

io/io.go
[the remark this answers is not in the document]
Koeng101: Gotcha, will do in the future. io/io.go has basically entirely died, so I'm fine with that here.

Usually git is able to figure out if I moved a file if I just immediately commit again

go.mod
```diff
@@ -15,8 +15,11 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/intel-go/cpuid v0.0.0-20181003105527-1a4a6f06a1c6 // indirect
```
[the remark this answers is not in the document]
Koeng101: I think it's just used in `StreamVByte`, but I don't think any other file format actually would benefit from this kind of compression very much. It is very nice in slow5 though, because of the integer data streaming out of the nanopore.

bio/bio.go
```diff
@@ -0,0 +1,275 @@
+/*
```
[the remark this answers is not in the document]
Koeng101: Let's definitely make an issue for that!

bio/genbank/genbank.go
```diff
@@ -1026,3 +1086,86 @@ func generateWhiteSpace(length int) string {
 GBK specific IO related things end here.
 
 ******************************************************************************/
+
+/******************************************************************************
+
+Old functions for testing here.
```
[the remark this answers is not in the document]
Koeng101: > If we are only keeping these for backwards test compatibility, why not just keep the capitalization?

Simplify the user API (for GoDocs and the such). We don't want people actually using those functions, and I'd prefer we get rid of them - it was just too much to do right now.

bio/genbank/genbank.go
```diff
@@ -33,7 +33,7 @@ GBK specific IO related things begin here.
 
 var (
 	readFileFn        = os.ReadFile
```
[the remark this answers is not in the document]
Koeng101: I guess it is a pattern, but I'm not a huge fan either. It's something I've just pulled along from older tests

bio/genbank/genbank.go
```diff
@@ -412,136 +449,159 @@ func (params *parseLoopParameters) init() {
...
+
+// WriteTo is a blank function, needed for compatibility with bio parsers. It doesn't do anything.
+func (header *Header) WriteTo(w io.Writer) (int64, error) {
+	return 0, nil
+}
+
+// Parser is a genbank parser created on an io.Reader.
+type Parser struct {
+	scanner    bufio.Scanner
+	parameters parseLoopParameters
```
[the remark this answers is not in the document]
Koeng101: Good point. The reason is basically historical, so I should fix that.

bio/fasta/fasta.go
```diff
@@ -0,0 +1,219 @@
...
+Many thanks to Jordan Campbell (https://github.com/0x106) for building the first
+parser for Poly and thanks to Tim Stiles (https://github.com/TimothyStiles)
+for helping complete that PR. This work expands on the previous work by allowing
+for concurrent  parsing and giving Poly a specific  parser subpackage,
+as well as few bug fixes.
+
+Fasta is a very simple file format for working with DNA, RNA, or protein sequences.
+It was first released in 1985 and is still widely used in bioinformatics.
+
+https://en.wikipedia.org/wiki/_format
```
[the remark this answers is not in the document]
Koeng101: Definitely, need to fix that

bio/fasta/fasta.go
```diff
@@ -0,0 +1,219 @@
...
+
+One interesting use of the concurrent  parser is working with the Uniprot
```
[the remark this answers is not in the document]
Koeng101: thank you for pointing that out

Koeng101:
TODO:
- Internalize parse parameters
- Add back fasta format in docs
- Remove that extra space

seqhash/seqhash_test.go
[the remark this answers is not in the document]
abondrn: Started thread in Discord

Koeng101:
> TODO:
> 
>     * Internalize parse parameters
> 
>     * Add back fasta format in docs
> 
>     * Remove that extra space

Parse parameters are actually there for reason. it's reset fairly often, so it is easier to have it outside the parser itself. We could have a separate method for fully resetting it, but I think it is fairly readable as is.

Koeng101:
Alright, looking over this it looks good to get a review from @TimothyStiles

TimothyStiles:
@Koeng101

I like the idea of utilizing generics but am not too familiar with their usage and syntax. Can you add a little "how-to" guide for writing parsers and writer in this new setup?

Koeng101:
> @Koeng101
> 
> I like the idea of utilizing generics but am not too familiar with their usage and syntax. Can you add a little "how-to" guide for writing parsers and writer in this new setup?

Hmmm, for both of those it doesn't really have much to do with generics (mostly just interfaces) - you only need generics once you make the top level convenience-function, which is essentially copy-pasted over each parser type, because Go generics are limited. Really what you have to be concerned about when writing parsers is the interfaces they must implement. To be clearer with the generics:

```go
// NewFastaParser initiates a new FASTA parser from an io.Reader.
func NewFastaParser(r io.Reader) (*Parser[*fasta.Record, *fasta.Header], error) {
	return NewFastaParserWithMaxLineLength(r, DefaultMaxLengths[Fasta])
}

// NewFastaParserWithMaxLineLength initiates a new FASTA parser from an
// io.Reader and a user-given maxLineLength.
func NewFastaParserWithMaxLineLength(r io.Reader, maxLineLength int) (*Parser[*fasta.Record, *fasta.Header], error) {
	return &Parser[*fasta.Record, *fasta.Header]{parserInterface: fasta.NewParser(r, maxLineLength)}, nil
}

// NewFastqParser initiates a new FASTQ parser from an io.Reader.
func NewFastqParser(r io.Reader) (*Parser[*fastq.Read, *fastq.Header], error) {
	return NewFastqParserWithMaxLineLength(r, DefaultMaxLengths[Fastq])
}

// NewFastqParserWithMaxLineLength initiates a new FASTQ parser from an
// io.Reader and a user-given maxLineLength.
func NewFastqParserWithMaxLineLength(r io.Reader, maxLineLength int) (*Parser[*fastq.Read, *fastq.Header], error) {
	return &Parser[*fastq.Read, *fastq.Header]{parserInterface: fastq.NewParser(r, maxLineLength)}, nil
}
```

What you can notice from both of these is that the Fastq is literally the same as the Fasta, except the `Data` and `Header` now points to fastq types rather than fasta types. The only reason you have to copy and paste this all the time instead of giving the format as a variable into a function is due to the limitation of Golang types.

## writers + parsers
"How to" make a parser is almost entirely contained within these few lines (as part of the type system):
```go
type parserInterface[Data io.WriterTo, Header io.WriterTo] interface {
	Header() (Header, error)
	Next() (Data, error)
}

type Parser[Data io.WriterTo, Header io.WriterTo] struct {
	parserInterface parserInterface[Data, Header]
}
```

Let's dive into the Golang type system to understand it better. A Parser is a generic struct, which means it is *any* struct that implements a `parserInterface`. What does that mean? Well, it means that any struct that implements the methods `Next()` and `Header()` satisfies the `parserInterface`. Both `Next()` and `Header()` just return an `io.WriterTo` and err, which is a standard library interface + primitive.

To sum it up, a `Parser` is *any* struct that returns a `io.WriterTo`+err when you call `Header()` or `Next()`. It is extremely simple. 

For example, the following parser would satisfy the `bio.Parser`:
```go
type SimpleHeader struct {}
type SimpleData struct {}
func (h SimpleHeader) WriteTo(w io.Writer) (n int64, err error) { return 0, nil}
func (d SimpleData) WriteTo(w io.Writer) (n int64, err error) { return 0, nil}

type SimpleParser struct{}
func (p SimpleParser) Header() (SimpleHeader, error) { return SimpleHeader{}, nil}
func (p SimpleParser) Next() (SimpleData, error) { return SimpleData{}, nil }
```

In this example, we have all the fundamentals needed:
1. A Header
2. A Data
3. An `io.WriteTo` implemented on the Header
4. An `io.WriteTo` implemented on the Data
5. A Parser
6. A `Header()` implemented on the parser
7. A `Next()` implemented on the parser

The clever bit is realizing that pretty much every single biological data structure has something along these lines, and that `Next()` supports stream-reading - and as a side effect of implementing the `io.WriteTo` interface, you also have writing built-in to the type definitions in a standard-library supported way.

In this way, the interfaces here are essentially just forcing parsers to expose 4 functions (`WriteTo` x2, `Next`, `Header`), and by doing that on their underlying data structures, you now can have a completely standardized way of working with each parser. 

## wrapping it up

Let's say you want to implement `SimpleParser` as a top level parser. Here is how you would do that:

```go
func NewSimpleParser() (*Parser[*SimpleData, *SimpleHeader], error) {
   return &Parser[*SimpleData, *SimpleHeader]{parserInterface: SimpleParser{}}, nil
}
```

All this does is:
1. Say that we are returning a Parser with the two types `SimpleData` and `SimpleHeader`
2. Implements the `parserInterface` needed for a Parser (because `SimpleParser` has those handy `Next()` and `Header()` functions)
3. returns the instantiated generic Parser, wrapping that underlying `SimpleParser`

## why we couldn't do this before generics

The hard bit is that I wanted to expose the underlying data types when you do things with a Parser. To pull from some above examples:

```go
parser, _ := bio.NewFastaParser(file)
fastaRecord, _ := parser.Next()
// you should be able to do the below!
fmt.Println(fastaRecord.Sequence)
fmt.Println(fastaRecord.Read)
```

Without a generic interface, we couldn't say that it was aight for `Data` to contain other things, like `Sequence` or `Name`, beyond the required `WriteTo`. Essentially, all the generics are doing is saying it is aight for `Data` to contain other things without having to go through complicated type conversion magic.

paleale:
Hello! Can I ask you to see and apply my changes for #352 ? 
I don't know a proper way for such changes to be suggested, so I've prepared a patch:
[0001-remove-not-working-error-handling-from-GetSequence-f.patch](https://github.com/TimothyStiles/poly/files/13464941/0001-remove-not-working-error-handling-from-GetSequence-f.patch)
Changes in CONTRIBUTION.md were discussed in #352 also... Thank you!

Koeng101:
> Hello! Can I ask you to see and apply my changes for #352 ? I don't know a proper way for such changes to be suggested, so I've prepared a patch: [0001-remove-not-working-error-handling-from-GetSequence-f.patch](https://github.com/TimothyStiles/poly/files/13464941/0001-remove-not-working-error-handling-from-GetSequence-f.patch) Changes in CONTRIBUTION.md were discussed in #352 also... Thank you!

Sure! Could you add a pull request with your changes into this branch? That would be the proper way + our tooling for reviewing changes would work well with that. Normally, you just make the changes on your own fork, and then do a pull request into this branch.

isaacguerreir:
It would be interesting now that we're using generics to have a factory to decide which parser to use based on the file content, so the developer doesn't need to choose what parser to use.

Koeng101:
> It would be interesting now that we're using generics to have a factory to decide which parser to use based on the file content, so the developer doesn't need to choose what parser to use.

Due to how generics work in Go, I'm not sure that you actually can do this. Basically, if the output signature is `Parser[Data, Header]`, you can't return `Parser[fasta.Name, fasta.Header]`.

isaacguerreir:
my mistake, for some reason reading your explanation I thought it would be possible to pass generic types as outputs, but taking a better look at the documentation it seems undoable to use Golang generics like that. It would be possible if the function returns a common structure, which is not possible with the current implementation.

Koeng101:
> seems undoable to use Golang generics like that. It would be possible if the function returns a common structure, which

Yes, we don't have a common structure between all of the different parsers, and I think this is a good thing. Makes each data type very specific and concise.

Issue #301, opened 2023-03-23
Certain tests fail on windows 


toothsy:
**Describe the bug**
Hello,I'm a windows user and I cloned the repo locally, installed all the vscode extensions, and ran all the tests, some test cases failed

**To Reproduce**
1. have a windows Operating system
2. clone the repo locally
3. run it in VSC
4. run all tests

**Expected behavior**
all tests must be green but some failed

**Screenshots**
![image](https://user-images.githubusercontent.com/48372345/227131575-65ac90ed-ea50-4084-a9fa-36d9822b4d16.png)

**Desktop:**
 - OS: [Windows]

**Additional context**
upon closer inspection in gff_test.go
- ![image](https://user-images.githubusercontent.com/48372345/227131786-4d4ba94a-4b08-4d10-8bce-502655e4e2fa.png)
     -  atoiFn is getting "6370\r" as its input hence it is panics
- the newline error due to windows is causing the issue

carreter:
Once #340 is merged, this will be an easy fix.

carreter:
Also @toothsy could you inspect the data being read in by the tests and let us know what the line endings are on Windows? You can do this using `git ls-files --eol`, please post the output of that here.

carreter:
Bumping this. If anyone has access to a Windows machine, info would be greatly appreciated! Worst case I can spin up a VM, but I'm lazy...

PR #340, opened 2023-09-08
Add a newline normalizing wrapper for io.Readers

carreter:
Will be useful for cross-platform support.

### DIFF
--- a/io/util.go
+++ b/io/util.go
@@ -1,0 +1,149 @@
+package io
+
+import "io"
+
+// trieNode is a simple implementation of a trie with values
+// of a single type. Only supports []byte keys for simplicity.
+type trieNode[V any] struct {
+	hasValue bool
+	value    V
+	children map[byte]*trieNode[V]
+}
+
+func (tn *trieNode[V]) Set(key []byte, val V) {
+	// Written iteratively rather than recursively for performance
+	// reasons and to avoid OOM errors.
+	currTn := tn
+	var nextTn *trieNode[V]
+	exists := false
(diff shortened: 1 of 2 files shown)


Koeng101:
Do you have any specific cases where this is useful? Is it like for, switching `\r` to `\n`?

carreter:
Yes, `\r` and `\r\n` to `\n`.

Koeng101:
Have you found any bioinformatics files that have `\r` and are so problematic?

carreter:
No, but I'm mostly thinking about making sure clients on MacOS, which IIRC we support, can parse files without worrying about normalizing the `\r` newlines to `\n` themselves.

Koeng101:
I see, so it would solve something like #301

carreter:
Hmm. I'm blanking on other ways to structure this that doesn't involve adding an intermediate. IMO we should wire this in in all the `NewXXXParserWithMaxLine` functions in `bio/bio.go`, this would minimally impact readability of the format-specific parsers themselves.

Koeng101:
We can't put it in `bio/bio.go` though because it would be a circular import, but I'm sure we could put it in some utils location

carreter:
I don't think there would be an import cycle. Regardless, I'm going to switch this to a draft PR until #339 is merged. Once it is, this wrapper con go in a new `ioutils` package:
```
bio/
├── util.go
└── specificFormat/
    ├── parser.go
    └── types.go
ioutils/
└── util.go
```

PR #341, opened 2023-09-11
Add svb compression to slow5

Koeng101:
Replaces #328 , merging this time with ioToBio instead to take advantage of the new parser format.


Koeng101:
Moving #328 here to merge with ioToBio.

https://github.com/TimothyStiles/poly/pull/328#discussion_r1310457760 is relevant - I think this is just looking ahead to blow5 compatibility. 

To quote the importance of just the svb compression:

>  I've seen reduction in total database size (storing fastq and slow5 reads) from 12Gb to 7.1Gb using just this one weird trick (database engineers hate him)

This is without implementing the other tricks from blow5.

carreter:
Continuing the thread from over there (https://github.com/TimothyStiles/poly/pull/328#discussion_r1321785254):

>  I'm using these functions for taking the rawSignal in and out of an SQL database, so having it be a method doesn't work as well as a raw function.
>
> CompressedRead I think would just be a blow5 read rather than slow5 read, which we SHOULD implement at some point. (there is a whole binary format specifically for doing those compressed reads well, with some real performance improvements)

If this is the case, I think we should entirely hold off on merging this PR in favor of doing a `blow5` PR instead. As it stands, I don't really see a use case for other clients of this package. If someone wants to perform this compression, they can just import your `svb` package themselves.

Koeng101:
> Continuing the thread from over there ([#328 (comment)](https://github.com/TimothyStiles/poly/pull/328#discussion_r1321785254)):
> 
> > I'm using these functions for taking the rawSignal in and out of an SQL database, so having it be a method doesn't work as well as a raw function.
> > CompressedRead I think would just be a blow5 read rather than slow5 read, which we SHOULD implement at some point. (there is a whole binary format specifically for doing those compressed reads well, with some real performance improvements)
> 
> If this is the case, I think we should entirely hold off on merging this PR in favor of doing a `blow5` PR instead. As it stands, I don't really see a use case for other clients of this package. If someone wants to perform this compression, they can just import your `svb` package themselves.


Breaking it down into two parts: 1 blow5 vs svb, and 2 svb triviality

I strongly disagree that blow5 fulfills the same use case:

1. `blow5` uses svb, but the use-case intention is completely different. These slow5 package functions are so that you can use slow5 with different data stores (like an SQL database). `blow5` cannot fulfill that use case, because you want to query on the raw data, so it doesn't exist as the binary format. However, the one column of data that takes the majority of space that won't be queried on directly can be compressed with svb to **halve** the needed storage, which is the limiting factor for these kinds of data files. `blow5` is for binary storage, probably putting the Nanopore runs in an object store, while slow5+svb is much better for a document-style database.
2. I am currently storing ~14 actual sequencing runs in a database with this format, with more coming in every week. I personally have real-world experience that this both works well in actual practice, and that `blow5` does not fulfill the niche. 

I also disagree that svb is trivial:

1. `svb` implementation is non-trivial (before I forked you couldn't actually do it with pre-existing libraries), and here we have it tested in the single context that it is useful. Unlike `gzip` or `zstd`, it isn't messing with an `io.Reader`/`io.Writer`. Trivial would be piping an `io.Reader`, not two `make([]xint)` plus a type conversion loop. What is a mask array? Do you need it? What inputs need to be saved for decompression? Having it there in a function increases readability of an unfamiliar compression/decompression type.
2. There is messaging with using svb in this way and having functions built-in to do it. The performance increase is pretty darn great compared to other compression methods, and by including it we are saying "THIS is how you should compress raw reads". 


> As it stands, I don't really see a use case for other clients of this package

I specifically talk about the [use-case for other clients in the context comment](https://github.com/TimothyStiles/poly/blob/f80b317632a475fca2676d86dcd2ae93b68eb1da/bio/slow5/slow5.go#L488-L489).

carreter:
Thank you for the clarifications re: slow5+svb vs blow5 and svb triviality. I completely missed the context comment, so sorry!

I'm on board with adding this functionality - reviewing now.

bio/slow5/slow5.go
```diff
@@ -473,3 +475,50 @@ func (read *Read) WriteTo(w io.Writer) (int64, error) {
...
+
+Stream VByte paper: https://doi.org/10.48550/arXiv.1709.08990
+
+Keoni
+
+******************************************************************************/
+
+// SvbCompressRawSignal takes a read and converts its raw signal field to two
+// arrays: a mask array and a data array. Both are needed for decompression.
+func SvbCompressRawSignal(rawSignal []int16) (mask, data []byte) {
```
carreter, line 505: I think it would be helpful to define a `CompressedRawSignal` struct to bundle `mask`, `data`, and `len(rawSignal)` as AFAICT, these data do not make sense independent of each other (? not sure, but this is what I'm guessing).

bio/slow5/slow5.go
```diff
@@ -473,3 +475,50 @@ func (read *Read) WriteTo(w io.Writer) (int64, error) {
...
+
+// SvbDecompressRawSignal decompresses raw signal back to a []int16. It
+// requires not only the mask array and data array returned by
+// SvbCompressRawSignal, but also the length of the raw signals.
+func SvbDecompressRawSignal(lenRawSignal int, mask, data []byte) []int16 {
+	rawSignalUint32 := make([]uint32, lenRawSignal)
+	rawSignal := make([]int16, lenRawSignal)
+	svb.Uint32Decode32(mask, data, rawSignalUint32)
+	for idx := 0; idx < lenRawSignal; idx++ {
+		rawSignal[idx] = int16(rawSignalUint32[idx])
```
carreter, line 521: Just to double-check: casting the `uint32` to an `int16` won't result in truncation here since raw signals are `int16`s, right?

bio/slow5/slow5_test.go
```diff
@@ -225,3 +225,32 @@ func TestSimpleExample(t *testing.T) {
...
+			break
+		}
+		outputReads = append(outputReads, *read)
+	}
+
+	for readNum, read := range outputReads {
+		rawSignal := read.RawSignal
+		mask, data := SvbCompressRawSignal(rawSignal)
+		rawSignalDecompressed := SvbDecompressRawSignal(len(rawSignal), mask, data)
+		for idx := range rawSignal {
```
carreter, line 250: Use `cmp.Diff` instead maybe?

carreter (changes requested):

bio/slow5/slow5.go
Koeng101: Yep, exactly. If you used the compression above on an int16, there should never be a time you encounter like an int32 (or equivalent) that would cause a truncation.

bio/slow5/slow5.go
Koeng101: I'm not sure actually. For additional context, all 3 (mask, data, len(rawSignal)) would actually be stored somewhere, so you would get them out separately (like in an SQL database, you'd have a `mask` column and a `data` column). At the same time, I could imagine if you want to store in a JSON format or a document store, it'd make way more sense to have all the numbers together, so you could just pass in a marshaled struct directly from the JSON without having to access the data... although you could just have that struct have a method to call Svb compress/decompress

Koeng101: Because of the IO nature of this, I'm just gonna keep it as a function for now.

PR #343, opened 2023-09-12
Re-enable golangci-lint.yml, turn on linters, and lint.

carreter:


io/pileup/pileup.go
```diff
@@ -49,11 +48,6 @@ import (
 
 // https://en.wikipedia.org/wiki/Pileup_format
 
-var (
-	gzipReaderFn = gzip.NewReader
```
TimothyStiles, line 53: I think we started doing this to make it easier to mock tests?

io/genbank/genbank.go
```diff
@@ -213,19 +211,12 @@ func WriteMulti(sequences []Genbank, path string) error {
 // Build builds a GBK byte slice to be written out to db or file.
 func Build(gbk Genbank) ([]byte, error) {
 	gbkSlice := []Genbank{gbk}
-	multiGBK, err := buildMultiNth(gbkSlice, -1)
+	multiGBK, err := BuildMulti(gbkSlice)
```
TimothyStiles, line 214: @Koeng101 wasn't `buildMultiNth` meant to deal with some annoying corner case?

io/pileup/pileup.go
Koeng101: I think the big changes to the parsers remove this anyway - since gzip gets tossed into generics, its easier to test

io/genbank/genbank.go
Koeng101: Hmmm not entirely sure, I don't quite remember

Koeng101: The build functions will mostly be moved into `bio` once we get there though, I just haven't started on genbank compatibility because it's gonna be quite sizable and deserves its own PR

carreter: The underlying logic is the same, it was just being buried a few function calls deep for some reason.

Before: `Build()` makes a single element slice and passes it to `BuildMulti()`, which immediately calls `buildMultiNth`. The `count` argument was unused, that's why I deleted it.

io/pileup/pileup.go
carreter: > I think we started doing this to make it easier to mock tests?

IMO, you shouldn't need to reach into a module and change its state to mock it, especially given how Go handles errors. We should favor onion architecture keep I/O as high up in the call chain as possible. This allows inner functions to be pure. The `gzipReaderFn` variable was unused anyway, that's why I removed it.

carreter: Go's interfaces are the solution for testing, IMO. If a function/object needs to perform I/O operations, it should take `io.Reader` and `io.Writer` interfaces as an argument/field. These interfaces can then be easily be mocked:

```
import "bufio"
import "bytes"

const someData = ";alskjdf;lkajsdlf"

func main() {
    var b bytes.Buffer
    mockedWriter := bufio.NewWriter(&b)
    FunctionThatTakesAWriter(mockedWriter)
    got := b.String()

    mockedReader := bytes.NewReader([]byte(someData))
    FunctionThatTakesAReader(mockedReader)
}
```

Koeng101: I agree here! Much more readable too. This is how the `bio` parsers work, and is how I mock test to completion `fasta` in ioToBio.

Koeng101 (approved):
Looks good! Mostly just style changes (obviously, since this turns on linter), and the non-style changes make sense.

TimothyStiles (approved):
Excellent pull request. Thanks @carreter for submitting and thanks @Koeng101 for reviewing!

MERGED

PR #344, opened 2023-09-15
created a super minimal mash function for sketching sequences.

TimothyStiles:
This package is meant to help create "sketches" via the mash sketching algorithm. There's plenty of room for optimization here but I've implemented the core idea.

Mash: fast genome and metagenome distance estimation using MinHash.

Ondov, B.D., Treangen, T.J., Melsted, P. et al.
Genome Biol 17, 132 (2016).
https://doi.org/10.1186/s13059-016-0997-x

Mash Screen: high-throughput sequence containment estimation for genome discovery.

Ondov, B., Starrett, G., Sappington, A. et al.
Genome Biol 20, 232 (2019).
https://doi.org/10.1186/s13059-019-1841-x


mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+Genome Biol 17, 132 (2016).
+https://doi.org/10.1186/s13059-016-0997-x
+
+Mash Screen: high-throughput sequence containment estimation for genome discovery.
+Ondov, B., Starrett, G., Sappington, A. et al.
+Genome Biol 20, 232 (2019).
+https://doi.org/10.1186/s13059-019-1841-x
+
+The idea is that we can sketch a sequence of data, and then compare the sketch to other sketches to see how similar they are.
+This saves a bunch of computation time and memory, because we don't have to compare the entire sequence to another sequence.
```
Koeng101, line 17: It'd be nice to have some context on how this differs from algorithms like minimap2. Computation upfront?

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+
+TTFN,
+Tim
+*/
+package mash
+
+import "github.com/spaolacci/murmur3"
+
+// sketch algorithm
```
Koeng101, line 26: It'd be nice to have more context for this. Where is the sketch algorithm?

Comments should go up to 80char

For this long of multiline, especially if just context, should have a multiline comment style

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+
+The idea is that we can sketch a sequence of data, and then compare the sketch to other sketches to see how similar they are.
+This saves a bunch of computation time and memory, because we don't have to compare the entire sequence to another sequence.
+
+TTFN,
+Tim
+*/
+package mash
+
+import "github.com/spaolacci/murmur3"
```
Koeng101, line 24: Why murmur3? (might be nice in top level comment)

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+// sketch algorithm
+// slide a window of size k along the sequence
+// for each kmer in the window, hash it to a 32 or 64 bit number
+// keep the minimum hash value of all the kmers in the window up to a given sketch size
+// the sketch is a vector of the minimum hash values
+
+// what are mash's inputs and outputs?
+// inputs: a
+
+type Mash struct {
```
Koeng101, line 35: I'd like a description now what a mash is

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+
+func NewMash(kmerSize uint, sketchSize uint) *Mash {
+	return &Mash{
+		KmerSize:   kmerSize,
+		SketchSize: sketchSize,
+		Sketches:   make([]uint32, sketchSize),
+	}
+}
+
+func (m *Mash) Sketch(sequence string) {
```
Koeng101, line 49: Is this the NewMash function? If so, a function description string would be nice

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+	// slide a window of size k along the sequence
+	for kmerStart := 0; kmerStart < len(sequence)-int(m.KmerSize); kmerStart++ {
```
Koeng101, line 51: for standardization sequence should be capitalized or lowercased

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+		kmer := sequence[kmerStart : kmerStart+int(m.KmerSize)]
+		// hash the kmer to a 32 bit number
+		hash := murmur3.Sum32([]byte(kmer))
+		// keep the minimum hash value of all the kmers in the window up to a given sketch size
+		// the sketch is a vector of the minimum hash values
+		var biggestHashIndex int
+
+		// find the biggest hash value in the sketch
```
Koeng101, line 59: OH THIS is how it works. Basically, you hash each kmer window, and then keep a list of hashes, small to large. The shared hashes are then compared. There should be a comment somewhere generally explaining the concept behind this in a bit more detail than just the sketches.

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+		for i := 0; i < len(m.Sketches); i++ {
```
Koeng101, line 60: Tim, using single letter variables? `i` instead of `idx` ? I'm shocked! (though I do like `i` more, it is more idiomatic)

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+				break
+			} else if m.Sketches[i] > m.Sketches[biggestHashIndex] {
+				biggestHashIndex = i
+			}
+		}
+		m.Sketches[biggestHashIndex] = hash
+	}
+}
+
+// distance algorithm
```
Koeng101, line 72: This function string isn't written properly, should be ``// Distance is a function that...``

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+// compare the sketches of two sequences
```
Koeng101, line 73: Lines should be up to 80char

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+				break
+			}
+		}
+	}
+	return 1 - (float64(sameHashes) / float64(len(m.Sketches)))
+}
+
+// for each kmer in the window, hash it to a 32 bit number
+// keep the minimum hash value of all the kmers in the window up to a given sketch size
+// the sketch is a vector of the minimum hash values
```
Koeng101, line 92: What's this?

mash/mash_test.go
```diff
@@ -0,0 +1,20 @@
...
+	fingerprint1.Sketch("ATGCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGA")
+
+	fingerprint2 := mash.NewMash(17, 10)
+	fingerprint2.Sketch("ATGCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGATCGA")
+
+	distance := fingerprint1.Distance(fingerprint2)
+	if distance != 0 {
+		t.Errorf("Expected distance to be 0, got %f", distance)
+	}
+}
```
Koeng101, line 20: No tests for affirming the algorithm actually works on more complicated sequences or in the null case

mash/mash.go
```diff
@@ -0,0 +1,92 @@
...
+		m.Sketches[biggestHashIndex] = hash
+	}
+}
+
+// distance algorithm
+// compare the sketches of two sequences
+// count the number of hashes that are the same
+// divide the number of hashes that are the same by the total number of hashes
+// the result is the distance between the two sequences
+func (m *Mash) Distance(other *Mash) float64 {
```
Koeng101, line 77: I personally think Distance should be a function (with two inputs) instead of a method, perhaps with a value instead of a pointer? Kinda like `strings.Contains()`

The reason is that I am imagining this being used across a LOT of data, and functional concurrency seems easier. I might be totally wrong here though - it could be that the performance gains with methods would be a lot better than what a purely functional approach would give you. 

Thoughts @carreter ?

mash/mash.go
TimothyStiles: Just a draft. Co-pilot really likes single letter variables.

mash/mash.go
TimothyStiles: Was choice used in original paper.

Wei Shen also had a benchmark test of non-cryptographic has functions in go and murmur3 was ranked second in speed.

Koeng101: Maybe link the benchmark in the in the documentation

mash/mash.go
Koeng101: It's almost like nearly everyone else using Golang uses `i` rather than `idx`

mash/mash.go
carreter: Do we want to turn on enforcement of this in the linter (`lll`)? I left it off because a lot of sequences in the tests are longgggg, but that could be fixed by moving them to a file or just using concatenation to build them.

mash/mash.go
carreter: Lol, I think `idx` is really only necessary if you're dealing with multiple indices on different objects - otherwise `i`, `j`, `k`, are fine.

mash/mash.go
carreter: I'm surprised the linter didn't catch this???

mash/mash.go
carreter: I think it's fine to have multiline comments that start each line with `//`, and IMO we should standardize this across the repo to be one way or the other - not a huge deal though.

mash/mash.go
carreter: Ah, this branch doesn't have my updates to the linter config merged. Can you rebase this to the tip of `main` before merging to ensure all the lint checks pass @TimothyStiles ?

mash/mash.go
Koeng101: TBH I only want it against comments, normal lines I'm fine with being `math. MaxInt`

mash/mash.go
carreter: I don't think there is a performance difference between methods and functions, totally unsure though.

I also don't think this makes functional concurrency harder, as methods are still first-class objects. Could easily do the following:

```
var center *Mash // The Mash we will use as the center for distance calculations
var others []*Mash // The Mashes we are comparing center to
distances := make([]float64, len(otherMashes))

for i, other := range others {
    // Reassignments necessary as we are using loop variables in closures.
    i := i
    other := other

    go func() {
        distances[i] = center.Distance(other)
    }
}
```

Imagine that I did a waitgroup or something in the above code to ensure all the goroutines complete before we try and read from `distances`.

carreter: Alternatively, imagine we had a `MapConcurrent` function:

```
func MapConcurrent[T any, O any](input []T, f func(T) O) []O {
     // Build output slice by applying function f to all elements of input, concurrently. 
}
```

Then you could simplify my previous example to this:

```
var center *Mash // The Mash we will use as the center for distance calculations
var others []*Mash // The Mashes we are comparing center to
distances := MapConcurrent(others, center.Distance)
```

carreter: I'm ambivalent as to leaving this as a method vs turning it into a function. I'm assuming distances are commutative - if they aren't, then I would have a slight preference for keeping it as a method to make the directionality clearer (`m` to `other`).

mash/mash.go
carreter: I'm on board, seems like I could do some regex magic to make that happen: https://github.com/walle/lll#usage. Will make an issue for this now.

mash/mash.go
Koeng101: You're right, it shouldn't matter. If I am reading it right, distances are commutative. I'm going to resolve!

Koeng101:
7zdkzz

TimothyStiles:
adding the help wanted tag because @Koeng101 wants this merged but the following still needs to be done and I won't be able to get around to it for at least a week.

1. Keep hashes sorted for faster search
2. Figure out where this belongs on the package level e.g `seqhash` or maybe a new `search` package?
3. Better comments and docs in general
4. 100% test coverage and more robust test case.

carreter:
Linking this to #356

mash/mash.go
```diff
@@ -0,0 +1,88 @@
...
+	var sameHashes int
+	for i := 0; i < len(m.Sketches); i++ {
+		for j := 0; j < len(other.Sketches); j++ {
```
soypat, line 80: This nested loop could be changed for two separate loops, one where the shorter sketch slice is stored into a `map[uint32]struct{}` type and the other where they are checked to see if other long uint32s slice contains entries which are in the map. Would speed up for large sketches.

soypat (commented):
I've suggested a radically different API for performance and future proof reasons. Let me know what y'all think.

Koeng101:
I really like @soypat changes in this case. They make a lot of sense to me

mash/mash.go
TimothyStiles: https://github.com/shenwei356/go-hashing-kmer-bench

mash/mash.go
TimothyStiles: fixed @carreter

mash/mash.go
TimothyStiles: > capitalized or lowercased

preference?

mash/mash.go
TimothyStiles: Yeah that's more or less how it works.

Keep it mind that sketches generated via different kmer window sizes are non-comparable. That's why I thought to store kmer size in the Mash struct @soypat because it could be relevant later. Open to better ways/ more idiomatic ways of doing this.

soypat: If Kmer size is important then changes I proposed are probably not the best way to go. Knowing this I'm thinking the following usage would be nice:

```go
m1, err :=mash.NewMash(seq1, MashConfig{Sketches: 23, KmerSize:k1})
// handle your errors...
m2, err := mash.NewMash(seq2,  MashConfig{Sketches: 23, KmerSize:k2})

 // Don't let users modify kmerSize field since it "invalidates" calculated data, do expose it via a method.
if m1.KmerSize() != m2.KmerSize() {
    panic("oh no!")
}

if mash.Equal(m1, m2) {
   print("mashes equal!")
}
```

mash/mash.go
```diff
@@ -0,0 +1,88 @@
...
+// what are mash's inputs and outputs?
+// inputs: a
+
+type Mash struct {
+	KmerSize   uint
+	SketchSize uint
+	Sketches   []uint32
+}
+
+func NewMash(kmerSize uint, sketchSize uint) *Mash {
```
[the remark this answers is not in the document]
soypat: Also I'd avoid `uint` in size definitions to catch programmer errors early. Having sizes as uint can do more harm than good i.e:

```go
sketchSize := uint(len(sketchPool) - totalProcessed)
```
In the code above the user calculates the sketch size from a pair of `int`s. If the user is mistaken and the result of the calculation is negative then `uint` converts the negative number into a massively huge number which would cause a out of memory panic.

Compared to negative size panics the OOM panic can affect other goroutines running in the same program.

mash/mash.go
TimothyStiles: @soypat think I'm going to try writing it like this today. Mashes are measured by their Jaccard Similarity (total intersection of hashes / actual hashes so instead will likely make it `mash.Simularity` and have it return a floating point between 0-1.

One other thing I can probably handle by myself is keeping the hashes organized by lexical least to lexically greatest during creation. Any advice on the fastest/most idiomatic way of doing this in Go?

mash/mash.go
Koeng101: "here is the science"

> I don't like the science

[marked this conversation as resolved.]

many such cases

mash/mash.go
TimothyStiles: @Koeng101 does the top level package comment resolve this?

mash/mash_test.go
TimothyStiles: Good point. Any ideas for a good test case @Koeng101?

Issue #367, opened 2023-09-23
Refactor `clone` package


carreter:
Per @Koeng101 's suggestion, the `clone` package needs some refactoring to make its code easier to understand and use.

Blocks #359 .

carreter:
Reopening this because the `clone` package still needs a good amount of work. Thinking of spending a significant amount of time on this during the first half of January, as most of the scripts I will be writing for Prof. Weiss' lab will be related to cloning.

@TwFlem would love to get your input on what you think needs to be done. You mentioned `seqhash` needing some optimization - I bet we could just look at recently published algos and make our own implementation. I'd also like to rewrite the structs that hold the data on DNA parts/fragments to make them more flexible and properly encode 3' vs 5' overhangs.

TwFlem:
@carreter I mentioned the seqhash performance in https://github.com/bebop/poly/pull/393 because the bug fix was related to concurrent code. Before removing it, I did some profiling to make sure it didn't cause any issues. It turns out removing the concurrency was a speedup. I only mentioned improving seqhash just in case if we needed cloning to be faster. 

Unless Poly consumers are currently raising concerns about performance, I would not say seqhash is fine the way it currently is.

carreter:
This is gonna be high-priority for me this January as I will be working to facilitate the cloning workflow of Fabio Caliendo in Prof. Ron Weiss' lab.

Currently blocks #359 .

PR #393, opened 2023-11-04
Cloning refactor and data race fix

### LINKED ISSUE #276: Data race causes flaky test in "clone"
**Describe the bug**

When running `clone` tests in a repeated way, `TestSignalKilledGoldenGate` randomly fails. I believe this might be caused by data races in the `clone` package, as reported by `go test -race ./clone`:
```
% go test -race ./clone
==================
WARNING: DATA RACE
Write at 0x00c0000ba0f0 by goroutine 45:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:272 +0x3c0
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Previous write at 0x00c0000ba0f0 by goroutine 40:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:272 +0x3c0
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Goroutine 45 (running) created at:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0x5f4
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Goroutine 40 (finished) created at:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0x5f4
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8
==================
==================
WARNING: DATA RACE
Read at 0x00c000268218 by goroutine 68:
  runtime.growslice()
      /usr/local/go/src/runtime/slice.go:178 +0x0
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:272 +0x39c
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Previous write at 0x00c000268218 by goroutine 34:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:272 +0x3c0
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Goroutine 68 (running) created at:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0x5f4
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8

Goroutine 34 (finished) created at:
  github.com/TimothyStiles/poly/clone.recurseLigate()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0x5f4
  github.com/TimothyStiles/poly/clone.recurseLigate.func2()
      /Users/matias/go/src/github.com/TimothyStiles/poly/clone/clone.go:273 +0xd8
==================
--- FAIL: TestSignalKilledGoldenGate (0.02s)
    clone_test.go:169: Should be only 4 looping sequences. Got: 7
    testing.go:1319: race detected during execution of test
FAIL
FAIL	github.com/TimothyStiles/poly/clone	1.317s
FAIL
```

**To Reproduce**
1. To run tests in a repeated way: `go test -count 10 -v ./clone/`
1. Then run the race detector: `go test -race ./clone` on the main directory.

**Expected behavior**
Data race should be fixed.

TwFlem:
## Changes in this PR
This is an attempt to start refactor cloning for issue #367. This will also probably address issue #276. 

One thing that I noticed in the clone package was all of the concurrency happening underneath golden gate. This made it a bit harder to grock what was going on. This also messed with the correctness of the program- hence #276. Perhaps more importantly, the unbounded number of go routines spinning up for pure, unblocking compute causes performance problems. So many go routines get created that the program begins spending most of its time thrashing back and forth between goroutines rather than just computing.

Here's some cpu profiles.

This is the cpu profile with the same benchmark in the changes in this PR with the concurrency that's currently in master: 
cpu original-concurrenty-10s

This is just with the currency removed:
cpu no-concurrency-10s

You'll see that with concurrency removed, more time is spent actually running golden gate rather than coordinating goroutines and gc time. In fact, the bottleneck looks to be seqhash. There is also a large amount of time on the GC. GC time could probably be improved by taking an iterative approach- that way there isn't too much pressure on the groutine's stack to keep track of everything. 

As far as what to make concurrent- I don't know enough about Poly of the field and how this actually gets used. If I had to throw out suggestions on where to best add some parallelism, it might be parallelizing multiple runs of GoldenGate or parallelize that for loop that calls to fan out the initial call to recurseLigate for each fragment and maybe synchronize updating whatever will keep track of existing seqhashes. Some discussion and experimentation could go a long ways here. 

Also it could be super valuable to optimize seqhash :man_shrugging: 

This PR also introduces the concept of EnzymeManager. The idea here is to abstract away that global enzyme map state so consumers could have a straight forward way of managing their enzyme data throughout the lifetime of the program. For example, they could add, remove, reorder, whatever Poly would want to have EnzymeManager implement. The main impetus for this was the comment

`Eventually, we want to get the data for this map from ftp://ftp.neb.com/pub/rebase`

Something like the EnzymeManager could make that flexibility possible and allow consumers to load whatever data they want from wherever they want. 

Honestly though, we can throw the EnzymeManager out- the other changes seem much more important. 

starts resolving #367 and will resolve #276. 

The EnzymeManager related changes are breaking. This isn't important for addressing #367 and #276- this can be removed if need be.

## Pre-merge checklist

* [x ] New packages/exported functions have docstrings.
* [x ] New/changed functionality is thoroughly tested. (the existing tests pass)
* [] New/changed functionality has a function giving an example of its usage in the associated test file. See `primers/primers_test.go` for what this might look like.
* [x ] All code is properly formatted and linted.
* [x ] The PR template is filled out.


Koeng101:
I was the original one to code clone, back when I wasn't very good at Go. I am sorry. Thank you for doing this.

> Here's some cpu profiles.

I need zero convincing to change this out of concurrency, lol

> Also it could be super valuable to optimize seqhash

I'm interested: how so? I'm not sure how we can check for equivalence, since boothLeastRotation is handling the circular sequences, where most time is spent.

> taking an iterative approach- that way there isn't too much pressure on the groutine's stack to keep track of everything.

This makes sense to me. I've always thought an iterative approach would be better.

> As far as what to make concurrent- I don't know enough about Poly of the field and how this actually gets used
 
This is somewhere I can help! I think I am the main user of this functionality, though I hope more folks will use it soon (and I use it to back some easier-to-use APIs for people). So, the idea is that I can use this for simulating GoldenGate assemblies that I am doing in the lab. 

The thing was, I made this functionality to truly model even the most complex GoldenGates. _I've literally never found someone who did these complexity GoldenGates_. So it solves a problem that doesn't really exist. Really what is needed is for the function to only simulate GoldenGates with ordered pairs of overhangs (so just checks that no two fragments share overhangs, then line them up, and put em together). Far simpler, and no need to even call seqhash.

The most complex use cases for combinatorial assembly can then simply iterate through each  "part type" and call the function. The most-most complex use case I built for (combinatorial and variant overhang fragments) should IMO just be removed. Nuke the code [not a request in this PR, but something I'd like to eventually do].

> Something like the EnzymeManager could make that flexibility possible and allow consumers to load whatever data they want from wherever they want.

I like the EnzymeManager.

Reviewing now.

clone/clone_test.go
```diff
@@ -1,33 +1,44 @@
...
 )
 
+// Eventually, we want to get the data for this map from ftp://ftp.neb.com/pub/rebase
+func getBaseRestrictionEnzymes() []Enzyme {
+	return []Enzyme{
+		{"BsaI", regexp.MustCompile("GGTCTC"), regexp.MustCompile("GAGACC"), 1, 4, "GGTCTC"},
+		{"BbsI", regexp.MustCompile("GAAGAC"), regexp.MustCompile("GTCTTC"), 2, 4, "GAAGAC"},
+		{"BtgZI", regexp.MustCompile("GCGATG"), regexp.MustCompile("CATCGC"), 10, 4, "GCGATG"},
+	}
+}
```
Koeng101, line 18: We should expose this to consumers. The number of enzymes actually used in GoldenGates is super limited. Tbh, I think the one change would be to have these as a nice default that users can use, as well as adding the following enzymes:

```go
    "PaqCI": {"PaqCI", regexp.MustCompile("CACCTGC"), regexp.MustCompile("GCAGGTG"), 4, 4, "CACCTGC"},
    "BsmBI": {"BsmBI", regexp.MustCompile("CGTCTC"), regexp.MustCompile("GAGACG"), 1, 4, "CGTCTC"},
    "SapI": {"SapI", regexp.MustCompile("GCTCTTC"), regexp.MustCompile("GAAGAGC"), 1, 3, "GCTCTTC"}
```

Koeng101 (approved):
One change request, which actually changes the usage of the package a little but is a feature I'd actually like to implement.

I like the changes though, and I think this is good for a merge.

clone/clone_test.go
TwFlem: Done!

TwFlem:
No problem and nothing to be sorry for! I was able to jump right in and reason about.

> > Also it could be super valuable to optimize seqhash
> 
> I'm interested: how so? ...

No clue! I'd have to take a closer look and understand what's going on under the hood. I only say this because that seemed to be a hotspot for Golden Gate. If performance becomes a problem- this might be a good place to start looking. 

> This makes sense to me. I've always thought an iterative approach would be better.

Yeah, I think this would be worth trying out and measuring. 

And ah- so maybe no point in trying to optimize this any further if it might get completely changed up.

Koeng101:
> I only say this because that seemed to be a hotspot for Golden Gate

I just wonder if there is anyway to optimize it (booth rotation algorithm is pretty good). Guessing best way is to just remove the need for it.

> And ah- so maybe no point in trying to optimize this any further if it might get completely changed up.

Yes, but this solves a race condition, so definitely worth a merge

Last thing: Can you add to the changelog? Then we can go for a merge. Thanks!

clone/clone.go
```diff
@@ -100,20 +111,14 @@ Base cloning functions begin here.
...
-
 // CutWithEnzymeByName cuts a given sequence with an enzyme represented by the
 // enzyme's name. It is a convenience wrapper around CutWithEnzyme that
 // allows us to specify the enzyme by name.
-func CutWithEnzymeByName(seq Part, directional bool, enzymeStr string) ([]Fragment, error) {
-	enzymeMap := getBaseRestrictionEnzymes()
-	if _, ok := enzymeMap[enzymeStr]; !ok {
-		return []Fragment{}, errors.New("Enzyme " + enzymeStr + " not found in enzymeMap")
+func (enzymeManager EnzymeManager) CutWithEnzymeByName(seq Part, directional bool, name string) ([]Fragment, error) {
+	if v, ok := enzymeManager.enzymeMap[name]; ok {
```
TimothyStiles, line 118: what is v here?

clone/clone.go
```diff
@@ -135,20 +143,20 @@ func CutWithEnzyme(seq Part, directional bool, enzyme Enzyme) []Fragment {
...
-			reverseOverhangs = append(reverseOverhangs, Overhang{Length: enzyme.OverhangLen, Position: reverseCut[0] - enzyme.Skip, Forward: false, RecognitionSitePlusSkipLength: len(enzyme.RecognitionSite) + enzyme.Skip})
+			reverseOverhangs = append(reverseOverhangs, Overhang{Length: enzyme.OverheadLength, Position: reverseCut[0] - enzyme.Skip, Forward: false, RecognitionSitePlusSkipLength: len(enzyme.RecognitionSite) + enzyme.Skip})
 		}
 	}
 
 	// If, on a linear sequence, the last overhang's position + EnzymeSkip + EnzymeOverhangLen is over the length of the sequence, remove that overhang.
 	for _, overhangSet := range [][]Overhang{forwardOverhangs, reverseOverhangs} {
 		if len(overhangSet) > 0 {
-			if !seq.Circular && (overhangSet[len(overhangSet)-1].Position+enzyme.Skip+enzyme.OverhangLen > len(sequence)) {
+			if !part.Circular && (overhangSet[len(overhangSet)-1].Position+enzyme.Skip+enzyme.OverheadLength > len(sequence)) {
```
TimothyStiles, line 159: could there be some explanation of the logic behind this line and what it's doing?

clone/clone_test.go
```diff
@@ -116,52 +111,40 @@ func TestCircularLigate(t *testing.T) {
...
 	if err == nil {
 		t.Errorf("GoldenGate should fail when using enzyme EcoRFake")
 	}
-	if err.Error() != "Enzyme EcoRFake not found in enzymeMap" {
+	if err.Error() != "Enzyme EcoRFake not found" {
 		t.Errorf("Failure of GoldenGate on incorrect enzyme should follow the exact string `Enzyme EcoRFake not found in enzymeMap`. Got: %s", err.Error())
 	}
 }
 
-func ExampleGoldenGate() {
```
TimothyStiles, line 132: @TwFlem we should have an example test for GoldenGate

clone/clone.go
```diff
@@ -333,14 +320,24 @@ Specific cloning functions begin here.
 
 // GoldenGate simulates a GoldenGate cloning reaction. As of right now, we only
 // support BsaI, BbsI, BtgZI, and BsmBI.
-func GoldenGate(sequences []Part, enzymeStr string) ([]string, []string, error) {
+func (enzymeManager EnzymeManager) GoldenGate(sequences []Part, enzymeStr string) (openConstructs []string, infiniteLoops []string, err error) {
```
TimothyStiles, line 323: should GoldenGate really be a method of enzymeManager rather than its function?

TwFlem: After learning more about how Golden Gate works IRL and a little bit more about SynBio- probably not. 

It seems more appropriate for Golden Gate to just accept something like a 'cuttingEnzyme' of type 'Enzyme' and have the caller lookup whatever enzyme they want.

This is also nice because now Golden Gate doesn't have to return an error anymore.

TwFlem: I renamed it to 'ExampleEnzymeManager_GoldenGate' for the linter- it's still there!

clone/clone_test.go
TwFlem: I renamed it to 'ExampleEnzymeManager_GoldenGate' for the linter- it's still there!

clone/clone.go
Koeng101: I think that is if the recognition sequence is at the very end of a circular DNA, and then cutting on the other side (ie, the beginning of the sequence).

clone/clone.go
Koeng101: I personally would like the change of GoldenGate to be its own function, independent of an enzymeManager.

TwFlem: Updated

clone/clone_test.go
TwFlem: Well- now it is back to ExampleGoldenGate :)

PR #401, opened 2023-11-15
Bump Golang version

matiasinsaurralde:
## Changes in this PR
Bumps Golang version to latest stable one:  [1.21.4](https://go.dev/dl/)

Also ran `go mod tidy`.

Just keeping it up to date and I also think there can be many benefits, for example I explored the new `slices` package which is recommended and more performant than `sort` methods in a few scenarios, etc.

No - tests are fine too

## Pre-merge checklist

* [ ] New packages/exported functions have docstrings.
* [ ] New/changed functionality is thoroughly tested.
* [ ] New/changed functionality has a function giving an example of its usage in the associated test file. See `primers/primers_test.go` for what this might look like.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/TimothyStiles/poly
 
-go 1.21.4
+go 1.21
 
 require (
 	github.com/google/go-cmp v0.5.8
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/TimothyStiles/poly
 
-go 1.18
+go 1.21.4
 
 require (
 	github.com/google/go-cmp v0.5.8
@@ -9,6 +9,7 @@ require (
 	github.com/mroth/weightedrand v0.4.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/sergi/go-diff v1.2.0
+	github.com/spaolacci/murmur3 v1.1.0
 	golang.org/x/exp v0.0.0-20230310171629-522b1b587ee0
 	lukechampine.com/blake3 v1.1.5
 )
@@ -16,7 +17,6 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/mattn/go-sqlite3 v1.14.13 // indirect
-	github.com/spaolacci/murmur3 v1.1.0 // indirect
 )
 
 require (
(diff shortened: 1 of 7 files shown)


matiasinsaurralde:
My only doubt here comes around c5a01d07c26b2e69002f0235e39a67c55fe9f27a, in theory it's properly using the `weightedrand` method to set a random source however tests fail. Would appreciate any feedback on it.

synthesis/codon/codon.go
```diff
@@ -149,11 +149,13 @@ func (table *TranslationTable) OptimizeSequence(aminoAcids string, randomState .
 	}
 
 	// weightedRand library insisted setting seed like this. Not sure what environmental side effects exist.
+	var randomSource rand.Source
 	if len(randomState) > 0 {
-		rand.Seed(int64(randomState[0]))
+		randomSource = rand.NewSource(int64(randomState[0]))
```
TimothyStiles, line 154: @matiasinsaurralde do you know if there's any change in behavior from the rand.Seed to rand.NewSource migration? This change seems to be breaking codon optimization tests.

carreter: This wasn't being used properly further down, fixed it in my latest commit.

go.mod
```diff
@@ -9,14 +9,14 @@ require (
 	github.com/mroth/weightedrand v0.4.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/sergi/go-diff v1.2.0
+	github.com/spaolacci/murmur3 v1.1.0
```
carreter, line 12: Why was this moved to be a direct dependency?

TimothyStiles:
@carreter if we're bumping the version then we should probably also bump the go versions that CI/CD runners use. Doing it now before merge.

carreter (approved):
LGTM except for single comment. `murmur3` probably should be in the indirect dependencies section.

go.mod
TimothyStiles: I'm not really sure @carreter. It's used only in the `mash` package similar to how blake3 is only used in the `seqhash` package.

TimothyStiles: Yeah had to double check but murmur3 should have been a direct dependency to begin with @carreter. Good to merge?

TimothyStiles (approved):
Looks good to me!

PR #408, opened 2023-12-01
fixed bug that produced wrong overhang in linear, non-directional, single cut reactions.

TimothyStiles:
## Changes in this PR
fixed bug that produced wrong overhangs in linear, non-directional, single cut reactions.

There was a bug in CutWithEnzyme where in the case of linear, non-directional, single cut reactions, it would assume that the overhangs were produced by a forward facing cut site and return a sequence downstream. In the case of the cut being a reverse cut site it would return the wrong overhangs. This was fixed with a simple if-else

There are no breaking changes.

## Pre-merge checklist

### DIFF
--- a/clone/clone.go
+++ b/clone/clone.go
@@ -188,7 +188,15 @@ func CutWithEnzyme(part Part, directional bool, enzyme Enzyme) []Fragment {
 		// In the case of a single cut in a linear sequence, we get two fragments with only 1 stick end
 		fragmentSequence1 := sequence[overhangs[0].Position+overhangs[0].Length:]
 		fragmentSequence2 := sequence[:overhangs[0].Position]
-		overhangSequence := sequence[overhangs[0].Position : overhangs[0].Position+overhangs[0].Length]
+
+		var overhangSequence string
+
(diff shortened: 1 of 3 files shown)


clone/clone.go
```diff
@@ -188,7 +188,15 @@ func CutWithEnzyme(part Part, directional bool, enzyme Enzyme) []Fragment {
 		// In the case of a single cut in a linear sequence, we get two fragments with only 1 stick end
 		fragmentSequence1 := sequence[overhangs[0].Position+overhangs[0].Length:]
 		fragmentSequence2 := sequence[:overhangs[0].Position]
-		overhangSequence := sequence[overhangs[0].Position : overhangs[0].Position+overhangs[0].Length]
+
+		var overhangSequence string
+
+		if len(forwardOverhangs) > 0 {
```
TimothyStiles, line 194: @Koeng101 should similar logic be implemented elsewhere in this function?

Koeng101:
@TimothyStiles Why did you nuke the changelog in d0f6ab3?

Issue #360, opened 2023-09-23
Implement Burrows-Wheeler transform


carreter:

carreter:
@TimothyStiles can you elaborate on what exactly it is we need here and triage the issue here + in the roadmap?

TwFlem:
I'd love to take a whack at this!

@TimothyStiles might have some other intentions in mind, but Burrows Wheeler Transform (BWT) can be used as a data structure to query whether or not a sub sequence within a sequence exists. 

This can be useful for determining whether or not a sub sequence of nucleotides exist within a sequence of nucleotides.

It is very memory efficient and exact matches on reads are fast.

There is also some room for backtracking on seeming mismatches  for some fuzzy sub sequence searching and maybe some other opportunities to tune the BWT like reporting the N number of locations where this sub sequence exists.

Maybe the exact case is a good first step? Do we want to report the location of the match as well- seems important?

Koeng101:
> There is also some room for backtracking on seeming mismatches for some fuzzy sub sequence searching and maybe some other opportunities to tune the BWT like reporting the N number of locations where this sub sequence exists.

On the alignment side, I think this basically becomes bwa https://bio-bwa.sourceforge.net . What I think could be neat is if you could somehow make it get 98% matching working, so it could be used for #396 in auto-annotating features. Right now that is done in plannotate with BLAST, but I'm pretty sure it is done that way because BLAST is an easy tool to just import and use. This could be a good application! 

For actual large-scale sequence alignment, minimap2 is probably a better route - tried and true with nanopore-type alignment, which IMO is the best upcoming DNA sequencing method. (Also it's what I use right now)

TimothyStiles:
> @TimothyStiles might have some other intentions in mind, but Burrows Wheeler Transform (BWT) can be used as a data structure to query whether or not a sub sequence within a sequence exists. 
> 
> This can be useful for determining whether or not a sub sequence of nucleotides exist within a sequence of nucleotides.
> 
> It is very memory efficient and exact matches on reads are fast.
> 
> There is also some room for backtracking on seeming mismatches  for some fuzzy sub sequence searching and maybe some other opportunities to tune the BWT like reporting the N number of locations where this sub sequence exists.
> 
> Maybe the exact case is a good first step? Do we want to report the location of the match as well- seems important?

@TwFlem I'd love to see you take a crack at this. At a conference on my phone rn.

[...]

PR #411, opened 2023-12-06
#360 Bwt

TwFlem:
## Changes in this PR

This is for #360. @TimothyStiles mentioned implementing a naive version of the BWT. At first I took his suggestion. As I read up on the applications of BWT for genomic data, the more it seemed important that we have as small as a memory footprint as possible. I then began to optimize Some things up front. And well... I ultimately followed Tim's suggestion about keeping it naive so we have something to start with :) 

Also, this does not include the fuzzy matching that @Koeng101. LMK if we want that in here, but that might be appropriate as its own thing since there's already a fair amount going on in this PR. 

There are a good chunk of auxiliary pieces to the BWT. I will talk about what was used in this PR along with some suggested ways we move forward to improve things. I will also point out the especially naive parts that most definitely need to be addressed. 

The BWT and other data structures added to enable alignment are documented in comments.

Things to Consider for the future:
1. Move towards using an RLFM and or r-index for significant space savings.
2. See if the Wavelet Tree would be faster if it were just a contiguous piece of memory
3. suffix array sampling- especially if we do 1. in the future
4.  Better select implementation for the RSA Bitvector

Tim hinted at a need for a way to compress sequences while still being able to analyze them in #360

This is all net new!

## Pre-merge checklist


Koeng101:
Very cool! Tell me when you need a code review

TimothyStiles:
Hey @TwFlem thanks for submitting this draft!

For naive I meant really, really naive. I'm not sure what everything here is doing yet but here's a simple 30 line BWT implementation that works:

```go
func burrowsWheelerTransform(input string) string {
	// Add a sentinel character to the end of the input
	input += "$"

	// Create a matrix to store rotations
	rotations := make([]string, len(input))
	for i := 0; i < len(input); i++ {
		rotations[i] = input[i:] + input[:i]
	}

	// Sort the rotations lexicographically
	sort.Strings(rotations)

	// Extract the last characters of each rotation to get the transformed string
	var result string
	for _, rotation := range rotations {
		result += string(rotation[len(rotation)-1])
	}

	return result
}
```

Obviously yours is more complex and very likely faster but I'm not sure what's going on yet. What's the best way to compare this^^ toy version with your own in this PR? What references did you use to create your draft? Any links?

-Tim

TwFlem:
Sure! So I read through many materials, but the most salient of them all came from Ben Langmead. He has a whole [youtube playlist](https://www.youtube.com/watch?v=5G2Db41pSHE&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&pp=iAQB) about the subject. 

I wish it wasn't so complex! So the BWT by itself is actually a small piece of the puzzle- there are auxiliary components that go along with the BWT to make it possible to both compress the size from the original string while also making it usable as a search index to perform other kinds of analysis like sequence alignment. 

The equivalent to what you wrote above is [here in the pr](https://github.com/TimothyStiles/poly/pull/411/files#diff-a57c1e2182129d8e4a533d106828ab4fd1b11e098828271bf3c3b882bab50cd1R116-R144) (ignoring the skiplist and the part about wavelet tree). The only difference with what I have here is that it saves some memory because we don't have to store NxN matrix of the string before we sort it.

Everything else in this PR is what is necessary to enable search and alignment. 

To sum up what I have in the PR and why it is there at a high level:
1. Bitvector- gives us a way to encode information in a memory efficient way. For example, instead of needing 1 byte to represent a character, all we need it 1 bit in the bitvector.
2. RSABitvector- RSA stands for Rank, Select, Access. So far, only Rank and Access are needed in the current implementation. However, select is needed for the RLFM/r-index implementations which would be the most usable implementation for anyone trying to do any serious work with the BWT. More on RLFM and r-index later though.
3. Wavelet tree- This is a data structure that allows us to make RSA queries on strings. TL;DR it is a binary tree that allows us to sub divide a given string and corresponding alphabet so we can make RSA queries. This is probably the most complex thing. I suggest just giving [this](https://www.youtube.com/watch?v=JszVzStky1E&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&index=9) a quick once over. 
4. LF mapping: LF stands for Last and First: Last column of the BWT transform and First column of the BWT transform. In this PR, F is the skiplist and L is just l for now. This is actually how we conduct search and alignment. In the Count, Locate, and Extract queries, you'll see we start with the whole range of both coulmns as the search range and use the skiplist along with ranking the L column to narrow down that search range. I suggest watching [this](https://www.youtube.com/watch?v=DYyXgxYmSYU&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&index=12) to see how the backwards search is conducted. It's a lot easier to see what's going on in the visual. 
5. prefixArray: This is a naive implementation. This takes up way too much memory as it currently is. The only reason why I didn't improve it further is because we'd most likely have to completely change its implementation if we want to do the RLFM and or r-index. The prefix array gives us a look up table that allows us map characterInTheFCoulmn -> characterInTheOriginalString. This allows us to extract the text. 

Also, I mentioned the RLFM and r-index. The only reason why I didn't continue with doing these is because this has gotten complex enough... that and I figured it'd be good to make sure you want all of this in the first place. With that being said, I highly recommend we pursue them. In a very similar manner to how the skipList in this PR is constructed, we can compress the L column substantially- especially for sequences that are very repetitive. He has some very compelling numbers in [this video](https://www.youtube.com/watch?v=Zf6r8XdYf-M&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&index=18) towards the end. Also, once you see the compression, it makes sense how much of an improvement this could be.  

Btw, I will fix the names and do some refactoring before taking this out of draft :).

bwt/bwt.go
```diff
@@ -0,0 +1,227 @@
+package bwt
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/exp/slices"
+)
+
+const nullChar = "0"
```
TimothyStiles, line 10: Is the null char similar to the "$" I've seen appended to the input string in other implementations?

bwt/bwt.go
```diff
@@ -0,0 +1,227 @@
...
+
+// BWT Burrow Wheeler Transform
```
TimothyStiles, line 12: I love this opening explanation. Makes it very clear why this is necessary.

bwt/bwt.go
TimothyStiles, line 10: Is the null char similar to the "$" I've seen appended to the input string in other implementations?

bwt/bwt.go
TimothyStiles: You've probably put this somewhere but I think the naive implementation as pseudocode may be good to explain the very simple core of the algo and help demystify it.

The stuff from Langmead should also be cited somewhere in the code itself.

bwt/bwt.go
```diff
@@ -0,0 +1,227 @@
...
+
+// getFCharPosFromOriginalSequenceCharPos looks up mapping from the original position
+// of the sequence to its corresponding posisiton in the First Column of the BWT
+func (bwt BWT) getFCharPosFromOriginalSequenceCharPos(originalPos int) int {
+	for i := range bwt.suffixArray {
+		if bwt.suffixArray[i] == originalPos {
+			return i
+		}
+	}
+	panic("Unable to find the corresponding orginal positiong for a character in the original sequence in the suffix array. This should not be possible and indicates a malformed BWT.")
```
TimothyStiles, line 89: this is a little meta but what are your thoughts on panics and when to use them? I almost always try to return an error and have it be handled upstream for fear of killing a user'ss long run.

bwt/bwt.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	sequence += nullChar
+
+	prefixArray := make([]string, len(sequence))
+	for i := 0; i < len(sequence); i++ {
+		prefixArray[i] = sequence[len(sequence)-i-1:]
+	}
+
+	// TODO: at the time of writing, the nullChar is 0, this is to ensure correctness in most cases.
```
TimothyStiles, line 174: I think this may be a good idea? Is this why I've been seeing "$" in other implementations?

bwt/bwt_test.go
```diff
@@ -0,0 +1,215 @@
...
+	fmt.Println(bwt.Count("CG"))
+	// Output: 10
+}
+
+type BWTCountTestCase struct {
+	seq      string
+	expected int
+}
+
+func TestBWT_Count(t *testing.T) {
```
TimothyStiles, line 29: Is this case just making a substring histogram?

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"golang.org/x/exp/slices"
+)
+
+// waveletTree datastructure that allows us to
+// conduct RSA queries on strings.
```
TimothyStiles, line 11: I know it's definitely spelled out somewhere in this PR but for newbies (and me!) it'd be good to have the abbreviation spelled out in parentheses the first time it appears in a file. What does RSA stand for again?

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+//
+// For the waveletTree's usage, please read the its
+// method documentation. To understand what it is and how
+// it works, then read below.
+//
+// # The WaveletTree has several imporant components
```
TimothyStiles, line 17: typo: `imporant` should be `important`

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+//
+// ## The Character's Path Encoding
+//
+// one important component is a character's path encoding.
+// which character we are woring with in a given path in the tree.
```
TimothyStiles, line 22: typo: `woring` should be `working`

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+//	     bananas
+//	    /       \
+//	  baaa      nns
+//	 /    \    /   \
+//	a      b  n     s
+//
+// ## RSA Intuition
+//
+// From here you may be able to build some intuition as to how we can take RSA queries given
+// a characters path encoding and which character we'd like to Rank, Select, and Access.
```
TimothyStiles, line 73: RSA is spelled out here! I was beginning to think I wouldn't find it.

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+// 5. child.Select(1, 1) of 0010101 is 4
+// 6. return 4 since we are at the root.
+//
+// ### Access Example
+//
+// If you've reached this point, then you must really be trying to understand how the
+// waveletTree works. I recommend thinking through how access could work with the example
+// above. HINT: it involves rank.
+//
+// NOTE: The waveletTree literlally have to be a tree. There are other forms that it may
```
TimothyStiles, line 100: typo: `literlally` should be `literally`

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+		} else {
+			curr = curr.left
+		}
+		level++
+		i = rank
+	}
+	return rank
+}
+
+// Select allows us to get the corresponding posisiton of a character
```
TimothyStiles, line 146: typo: `posisiton` should be `position`

bwt/wavelet.go
```diff
@@ -0,0 +1,340 @@
...
+		level++
+	}
+
+	for curr.parent != nil {
+		curr = curr.parent
+		level--
+		pathBit := ci.path.getBit(ci.path.len() - 1 - level)
+		rank, ok := curr.data.Select(pathBit, rank)
+		if !ok {
+			msg := fmt.Sprintf("could not find a correspodning bit for node.Select(%t, %d) for characterInfo %+v", pathBit, rank, ci)
```
TimothyStiles, line 169: typo: `correspodning` should be `corresponding`

bwt/bwt.go
TwFlem: For sure! I'll do something similar with what I did with the wavelet tree to illustrate things like:
* the BWT itself
* how the first and last column have characters of the same rank in the same order that enables all the magic
* LF mapping
* Compressing "runny" sequences like the first column.

I'd be happy to cite him. Are there any examples of citation that exist in Poly? If not, is there a preferred way to do that?

bwt/bwt.go
TwFlem: Yes exactly. It was "$" at first. I made it 0 for now because lexicographically, that's the character with the lowest value. This is just so if someone were to try and use the BWT with a string with spaces in it just to mess around and understand how it works, then it would work for them. Some follow up work that should be done:
1. modify the sorting so if we encounter the `nullChar` we always make sure that is the least. Sort everything else lexographically. I can give this a shot in this PR if you want though. I didn't want to add more noise than necessary.
2. Make that nullChar configurable. Neither "$" or "0" should show up in the protein, dna, or rna alphabets but... you never know how users might want to use this.

bwt/bwt.go
TwFlem: It depends. I'm usually very hesitant to put them in anywhere as well. However in these cases in the PR, the panics are analogous to having an index out of bounds error for an array. If something wrong is happening like this in the internals of any of these data structures, then something has gone horribly wrong and the program can't recover since the BWT is likely borked. 

That or if whoever is using is upstream is using it incorrectly, they can't recover from that incorrectness. Like in the case of the panic on the out of bounds on Extract. Imagine having to handle errors from index out of bounds.

bwt/bwt.go
TwFlem: I think I'll just give the custom sorting a shot on the string to ensure that the nullChar is always the least no matter what so we can put back the $.

bwt/bwt_test.go
TwFlem: I'm actually not sure what Count could be used for in the real world tbh.

I'm thinking that it could be a quick way to check for the existence of an exact match? That or for statistics like you mentioned?

Might be a good question for the community- is this useful functionality to expose?

bwt/wavelet.go
TwFlem: Done. Also, I'll definitely be fixing all the others because of the lint.

With that being said, I'm not breaking Poly's bank when a push up these small commits am I :thinking: ? I can be more considerate if that is the case.

bwt/wavelet.go
TwFlem: Sorry about that. Yeah, I was thinking that was more appropriate to spell out in the RSABitvector and just telling readers to look there to learn more about RSA. Happy to make that redundant though so maintainers don't have to jump around to Grok all of these concepts.

TwFlem:
> It'd be cool to do a little talk and demo for this.

I'd be happy to do that. Right now, there's not too much to demo or talk about maybe. We can talk about saving memory, indexing strings, the magic of the BWT and how rank is at the center of that magic.

Maybe it'd be even more valuable to have a demo once we have the fuzzy matching that Koeng101 mentioned in #360 along with having the RLFM and r-index? At that point, the BWT seems like it would be feature complete enough to reach a critical mass of usefulness for Poly's users.  

> It's a great example for teaching and maybe we should share it back with Ben? Just some late night thoughts.

I'd definitely like to at least thank him! It's amazing that he is giving that knowledge to the community like that. Perhaps his channel deserves a spot in https://github.com/TimothyStiles/how-to-synbio.

bwt/bwt.go
TwFlem: Thought about this some more... Also, sounds like the Go team has some opinions on this: https://go.dev/blog/defer-panic-and-recover

It does seem weird to bubble up errors like this because it is an internal problem, but Poly is also the first open source project I've contributed to- perhaps its best to give the users of Poly the choice of how they want to handle potential errors.

Maybe the example of the json package they mentioned in the blog post would also be appropriate here. We can panic internally and recover at the API boundary. That or we could just bubble errors all the way up for safety if for some reason we ever wanted to reuse the auxiliary components in here so we don't have any hidden panics.

Happy to change it up!

bwt/wavelet.go
TwFlem: fixed

bwt/wavelet.go
TwFlem: fixed

bwt/wavelet.go
TwFlem: I put Definitions of Acronyms towards the top of explanations like this. LMK if that works out.

bwt/wavelet.go
TwFlem: done

bwt/wavelet.go
TwFlem: Addressed, LMK if that's better.

bwt/bwt_test.go
TwFlem: @Koeng101 is the Count of sub-sequences useful to Expose?

bwt/bwt.go
TwFlem: Done, LMK what you think. nullChar could be configurable as a functional option now if users need to have $ as a part of their alphabet for the index they are throwing in to the BWT for whatever reason.

bwt/bwt.go
TwFlem: Still happy to change it up. With that said, I think the panics in bitvector are appropriate.

bwt/bwt.go
TwFlem: It can be whatever we want now! Replied to your other comment.

bwt/bwt_test.go
Koeng101: Yes, it is! There are instances where you are looking for multiple counts of sub-sequences (like restriction enzyme sites). Generally speaking, useful to have counts.

bwt/bwt.go
TwFlem: Started changing it up in f5b459b07749bf1b4f45f7f2d034d23cd97aaf5b. LMK What you think.

carreter:
What's the status of this? Ready for review?

TwFlem:
@carreter Yes it is!

carreter:
I'll take a look at this tomorrow then!

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+B: 001
+C: 010
+D: 011
+E: 100
+F: 101
+G: 110
+H: 111
+
+If we wanted to get to the leaf that represent the character D, we'd
+take the path:
```
carreter, line 38:
```suggestion
take the path that corresponds to the character's encoding, considering a 0 as choosing the left
child of a node and a 1 as choosing the right child of a node:
```
For clarity

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+
+For the waveletTree's usage, please read the its
+method documentation. To understand what it is and how
+it works for either curiosity or maintenance, then read below.
+
+# WaveletTree
+
+The Wavelet Tree allows us to conduct RSA queries on strings. in
+a memory and run time efficient manner.
+RSA stands for (R)ank, (S)elect, (A)ccess.
```
carreter, line 20:
```suggestion
RSA stands for (R)ank, (S)elect, (A)ccess.

See this blog post by Alex Bowe for an additional explanation:
https://www.alexbowe.com/wavelet-trees/
```
I found this explanation's visuals very helpful, I'd recommend including it.

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+It has the alphabet b, a, n, s
+Let's say it has the encoding:
+a: 00
+n: 01
+b: 10
+s: 11
+and that 0 if left and 1 is right
+We can represent this tree with bitvectors:
+
+     0010101
```
carreter, line 60:
```suggestion
     0010101
     bananas
```
I think including the string translation of the bitvector as Alex does in [this blog post](https://www.alexbowe.com/wavelet-trees/) would be very helpful!

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+## The Data Represented at each node
+
+Let us consider the sequence "bananas"
+It has the alphabet b, a, n, s
+Let's say it has the encoding:
+a: 00
+n: 01
+b: 10
+s: 11
+and that 0 if left and 1 is right
```
carreter, line 57:
```suggestion
and that 0 is left and 1 is right
```
nit, typo

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+a characters path encoding and which character we'd like to Rank, Select, and Access.
+
+### Rank Example
+
+To get WaveletTree.Rank(a, 4) of bananas where a's encoding is 00
+1. root.Rank(0, 4) of 0010101 is 2
+2. Visit Left Child
+3. child.Rank(0, 2) of 1000 is 1
+4. Visit Left Child
+5. return 1
```
carreter, line 86:
```suggestion
5. child.Rank(0, n) of a leaf node is simply n, so here return 1
```
For clarity

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+ /    \    /   \
+a      b  n     s
+
+## RSA Intuition
+
+From here you may be able to build some intuition as to how we can take RSA queries given
+a characters path encoding and which character we'd like to Rank, Select, and Access.
+
+### Rank Example
+
```
carreter, line 80:
```suggestion

WaveletTree.Rank(c, n) returns the rank of character c at index n in a sequence, i.e. how many
times c has occurred in a sequence before index n.

```

Just adding definition of rank because I didn't know it myself.

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+a      n  b     s
+
+If we translate each bit vector to its corresponding string, then it becomes:
+
+     bananas
+    /       \
+  baaa      nns
+ /    \    /   \
+a      b  n     s
+
```
carreter, line 73:
```suggestion

Each node of the tree consists of a bitvector whose values indicate whether
the character at a particular index is in the left (0) or right (1) child of the
tree.

```
Adding bit more information for clarity.

bwt/wavelet.go
```diff
@@ -0,0 +1,356 @@
...
+3. parent.Select(0, 1) of 001 is 1
+4. Go to the next parent
+5. parent.Select(1, 1) of 0010101 is 4
+6. return 4 since we are at the root.
+
+### Access Example
+
+If you've reached this point, then you must really be trying to understand how the
+waveletTree works. I recommend thinking through how access could work with the example
+above. HINT: rank might help.
```
carreter, line 102: I think we should explain it explicity anyways for ease-of-use.

```suggestion
above. HINT: rank might help.

Take the tree we constructed earlier to represent the sequence "bananas".

     0010101
    /       \
  1000      001
 /    \    /   \
a      n  b     s

To access the 4th character of the sequence, we would call WaveletTree.Access(3),
which performs the following operations:
1. root[3] is 0 and root.Rank(0, 3) is 2
2. Since root[3] is 0, visit left child
3. child[2] is 0 and child.Rank(0, 2) is 1
4. Since child[2] is 0, visit left child
5. Left child is a leaf, so we've found our value (a)!

In pseudocode:

def f(node, index):
    if node is leaf:
        return character
    
    val := node[index]
    rank := node.Rank(val, index)

    nextNode := node.leftChild if val is 0, node.rightChild if val is 1
    return f(nextNode, rank)

```

carreter (commented):
I've only had the chance to look at the docstrings and try and wrap my head around the high-level structure of things, will go back and look at the code itself later.

carreter (commented):
I've only had the chance to look at the docstrings and try and wrap my head around the high-level structure of things, will go back and look at the code itself later.

bwt/bwt.go
TwFlem: I Cited him at the end of the BWT's top comment.

bwt/bwt.go
```diff
@@ -0,0 +1,424 @@
...
+// to reduce its memory footprint while also maintaining a way
+// to analyze and work with the sequence.
+type BWT struct {
+	// firstColumnSkipList is the first column of the BWT. It is
+	// represented as a list of skipEntries because the first column of
+	// the BWT is always lexicographically ordered. This saves time and memory.
+	firstColumnSkipList []skipEntry
+	// Column last column of the BWT- the actual textual representation
+	// of the BWT.
+	lastCoulmn waveletTree
```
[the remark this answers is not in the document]
TwFlem: Updated.

bwt/bwt.go
```diff
@@ -0,0 +1,424 @@
...
+	// suffixArray an array that allows us to map a position in the first
+	// column to a position in the original sequence. This is needed to be
+	// able to extract text from the BWT.
+	suffixArray []int
+}
+
+// Count represents the number of times the provided pattern
+// shows up in the original sequence.
+func (bwt BWT) Count(pattern string) (count int, err error) {
+	defer bwtRecovery("Count", &err)
```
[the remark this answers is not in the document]
TwFlem: Spoke a bit about it [here](https://github.com/bebop/poly/pull/411#discussion_r1421267162)

There are some panics that happen in the non public facing parts of the BWT that didn't seem appropriate to bubble up like trying to access some index that it out of bounds on a bitvector etc. 

That Go blog post mentioned what the encoding/json package does internally. It seems important to give users the choice on how to handle the error- especially since they could be doing something very important or time consuming like Tim mentioned.

bwt/wavelet.go
TwFlem: I updated the explanation like you suggested, but changed the wording a bit.

bwt/wavelet.go
TwFlem: Fair. I'll definitely put in that explanation. 

The psuedocode could be redundant since the explanation is there along with the implementation below that could be used as a reference.

bwt/bitvector_test.go
```diff
@@ -0,0 +1,112 @@
+package bwt
+
+import (
+	"testing"
+)
+
+type GetBitTestCase struct {
```
TimothyStiles, line 7: should this be exposed?

bwt/bwt_test.go
```diff
@@ -0,0 +1,265 @@
...
+import (
+	"fmt"
+	"log"
+	"strings"
+	"testing"
+
+	"golang.org/x/exp/slices"
+)
+
+func ExampleBWT_Count() {
```
TimothyStiles, line 12: Example tests should be in their own example_test.go file with the package name `bwt_test` so that package functions are called like `bwt.Count` to more closely emulate usage outside the package.

bwt/bwt_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	count, err := bwt.Count("CG")
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(count)
+	// Output: 10
+}
+
+type BWTCountTestCase struct {
```
TimothyStiles, line 28: should this be exposed?

bwt/bwt_test.go
```diff
@@ -0,0 +1,265 @@
...
+		if err != nil {
+			t.Fatalf("seq=%s unexpectedError=%s", v.seq, err)
+		}
+		if count != v.expected {
+			t.Fatalf("seq=%s expectedCount=%v actualCount=%v", v.seq, v.expected, count)
+		}
+	}
+}
+
+func ExampleBWT_Locate() {
```
TimothyStiles, line 66: in example_test.go file?

bwt/bwt_test.go
```diff
@@ -0,0 +1,265 @@
...
+				t.Fatalf("seq=%s expectedOffsets=%v actualOffsets=%v", v.seq, v.expected, offsets)
+			}
+		}
+	}
+}
+
+func ExampleBWT_Extract() {
+	inputSequence := "AACCTGCCGTCGGGGCTGCCCGTCGCGGGACGTCGAAACGTGGGGCGAAACGTG"
+
+	bwt, err := New(inputSequence)
```
TimothyStiles, line 130: move to example_test.go

bwt/bwt_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	extracted, err := bwt.Extract(48, 54)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(extracted)
+	// Output: AACGTG
+}
+
+type BWTExtractTestCase struct {
```
TimothyStiles, line 143: should this be exposed?

bwt/bwt.go
```diff
@@ -0,0 +1,427 @@
...
+
+	searchRange := bwt.lfSearch(pattern)
+	return searchRange.end - searchRange.start, nil
+}
+
+// Locate returns a list of offsets at which the begging
```
TimothyStiles, line 206: typo: begging should be beginning?

bwt/wavelet.go
```diff
@@ -0,0 +1,381 @@
...
+	return n.char != 0
+}
+
+type charInfo struct {
+	char    byte
+	maxRank int
+	path    bitvector
+}
+
+func NewWaveletTreeFromString(str string) waveletTree {
```
TimothyStiles, line 246: is this supposed to be public? Needs Doc string

TimothyStiles (commented):
@TwFlem this is looking really good!

Exposed test structs should be removed and example tests should be in their own separate `example_test.go` file in a separate `bwt_test package`.

It's not clear where a lot of the test cases are coming from and there should be more comments about where they're from, and what corner case their testing.

I'm with @carreter that it'd be good to wrap my head around the actual concepts but I'm also thinking we should find some outside folk more familiar with bwt to review this too if we can.

TwFlem:
> It's not clear where a lot of the test cases are coming from and there should be more comments about where they're from, and what corner case their testing.

I added some comments about some edge cases that are being tested like with Jacobson's rank and the rsa_bitvector rank tests. I added a mention of how useful the the reconstruction test is for the WaveletTree. I also added a reconstruction test that pretty much guarantees that whatever we use for the suffix array in the future is well formed- otherwise we would never be able to reconstruct the string.

I added some more tests and more validation. So now we throw errors for empty and other invalid inputs. I fixed an edge case where a user could supply a string with a single alphabet character. 

Otherwise, there aren't really too many edge cases to go around with the FM index BWT. It's all predicated on the fact that the LF search works. LF search is all predicated on the fact that the Wavelet Tree and other auxiliary DSAs work as expected. The other tests are just to meant the show that auxiliary pieces work as expected. The RSA and Bit vector test cases are verbose because of all the bit math involved. Those are there to give us peace of mind that it works as expected. 

There will be some more BWT edge cases with the other improvements that I'll follow up with like the RLFM index (done) and the r-index (in progress). Those I can definitely call out. 

> I'm with @carreter that it'd be good to wrap my head around the actual concepts but I'm also thinking we should find some outside folk more familiar with bwt to review this too if we can. Those changes will be pretty small btw. Nothing like getting the basic BWT off the ground :sob: 

Sounds good! It might be better to emphasize the correctness over performance improvements. Reasonable performance improvements are already on the way :eyes:

bwt/wavelet.go
TwFlem: It is not. ty. Made it private.

bwt/bwt.go
TwFlem: done

bwt/bitvector_test.go
TwFlem: It Shouldn't be. Definitions in _test aren't public. 

```go
package poly

import (
	"fmt"

	"github.com/bebop/poly/bwt"
)

func polyRoot() {
	thing := bwt.GetBitTestCase{}
	fmt.Println(thing)
}

```

^ That won't build

bwt/bwt_test.go
TwFlem: Done

bwt/bwt_test.go
TwFlem: It shouldn't be. Explained Above.

bwt/bwt_test.go
TwFlem: Done

bwt/bwt_test.go
TwFlem: Done

bwt/bwt_test.go
TwFlem: It shouldn't be. Explained above.

bwt/bwt.go
```diff
@@ -0,0 +1,476 @@
...
+3. Repeat 2. N times where N is the length of the sequence
+4. Lexicographically sort the NxN matrix of rotated sequences where
+   the null termination character is always the least-valued
+5. Store the first and last column of the matrix. The last column
+   is the output of the BWT. The first column is needed to run queries
+   on the BWT of the original sequence.
+
+Lets use banana as an example.
+
+banana$      $banana
```
TimothyStiles, line 42: can there be an example test that takes `banana$` and returns the last column as a string?

TwFlem: @TimothyStiles updated

bwt/bwt.go
```diff
@@ -0,0 +1,481 @@
...
+		msg := fmt.Sprintf("offset [%d] exceeds the min bound of the BWT [0]", offset)
+		panic(msg)
+	}
+
+	for i := range bwt.firstColumnSkipList {
+		if bwt.firstColumnSkipList[i].openEndedInterval.start <= offset && offset < bwt.firstColumnSkipList[i].openEndedInterval.end {
+			return bwt.firstColumnSkipList[i]
+		}
+	}
+	panic("figure out what to do here")
```
TimothyStiles, line 336: what should be done here?

bwt/example_test.go
```diff
@@ -0,0 +1,70 @@
+package bwt_test
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/bebop/poly/bwt"
+	"golang.org/x/exp/slices"
+)
+
```
TimothyStiles, line 10: should have an Example_basic test here showing how to use the package's basic functionality end-to-end

TwFlem: Updated https://github.com/bebop/poly/pull/411/commits/1cc37557ed1da0cd9b1ef350466d5844ef375f11

bwt/bwt.go
TwFlem: Ty, updated this along with one hiding in the bitvector https://github.com/bebop/poly/pull/411/commits/1cc37557ed1da0cd9b1ef350466d5844ef375f11

Issue #434, opened 2023-12-20
Genbank parser needs heavy refactor or a rewrite


carreter:
## Describe the desired feature/enhancement
The `io/genbank` package needs a significant refactor or to be re-written entirely.

### Is your feature request related to a problem?
There are several outstanding issues with the parser that will be difficult to address in its current state: #383 #352 #351 #342 

### Describe alternatives you've considered (optional)
Main alternative is to hack over the code as-is to fix bugs, but I think this will take more time than a rewrite.

carreter:
Dropping a link to the GenBank spec here: https://www.ncbi.nlm.nih.gov/genbank/release/current/

TwFlem:
That's Neat format :thinking: 

Happy to take a whack at a rewrite. 

A few questions:
1. Do we want to incrementally rewrite this? For instance, starting with just parsing a genbank file? Would a good place to start be with just the info in the header? This way, we can discuss what error handling looks like, how we want things organized, setting the tone for how we want to parse the rest of the info, etc.
2. Do we want to start with the latest version of the spec? Or do we want to build out the backwards compatibility for spec breaking changes along the way?

carreter:
1. I actually just kinda went at it, but would be happy to take a step back and replan - I've definitely noticed some weaknesses in my approach. I'll clean up the code and get it in committable state for just the header!
2. I was targeting the latest version of the spec in my approach, but unsure what the best approach is of the two options you present.

TwFlem:
re 1.

Oh nice! I don't want to step on your toes- so no pressure. Maybe there's room to divy up different parts of the spec after establishing how the header will be parsed? If so, I'd be happy to help with that! 

re 2.

I think targeting the latest version of the spec is definitely the better option. Adding backwards compat after the fact seems straight forward in this case. It doesn't look like there are too many breaking changes and they seem pretty separate from everything else.

carreter:
Not stepping on any toes :) I'll try and get that code up today, I'm away from the PC I had the code on.

Feel free to take a stab at any issues you would like!!!

carreter:
Ok, draft PR is up that has the types + most of the layout of the parsers logic. @TwFlem let me know what you think!

TwFlem:
@carreter That looks super good! That pattern will go along with way the genbank parser along with the other parsers as well. The error type looks super good as well. It looks flexible enough so we can get nice and specific about what it wrong. The way things are being done just looks good in general.

The only real value added suggestions are about testing.

The others are more speculation.

PR #437, opened 2024-01-11
Genbank parser rewrite

carreter:
## Changes in this PR
Rewrite of the Genbank parser. Currently in a draft state.

See #434 .

Yes. Changes to the structure and methods of the `Genbank` struct, as well as the API for the parser.

## Pre-merge checklist

* [ ] New packages/exported functions have docstrings.
* [ ] New/changed functionality is thoroughly tested.
* [ ] New/changed functionality has a function giving an example of its usage in the associated test file. See `primers/primers_test.go` for what this might look like.
* [ ] The PR template is filled out.


io/genbank/parser_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+func TestParseHeader(t *testing.T) {
+	testcases := []struct {
+		name    string
+		data    string
+		want    Header
+		wantErr bool
+	}{{
+		name: "parses example header from spec",
+		data: `GBBCT1.SEQ          Genetic Sequence Data Bank
```
TwFlem, line 19: For the cases where `wantErr` is nil, perhaps it could be helpful to define `data` in terms of want. 

So for instance, `Header` could have a method like

```go
func (h Header) bytes() []bytes {
  strbuilder := strings.Builder{}
  for _, r := range h.FileName {
     strbuilder.WriteRune(r)
  }
  for i := 0; i < paddingBeforeTitle-len(h.FileName); i++ {
      strbuilder.WriteByte(" "[0]) // maybe hard code some common byte values for these :shrug: 
   }
  for _, r := range h.Title {
     strbuilder.WriteRune(r)
  }
  strbuilder.WriteByte("\n"[0])
  // etc.
}


	for _, tc := range testcases {
		t.Run(tc.name, func(t *testing.T) {
                        var p *Parser
                        if data == "" {
                            p = NewParser(strings.NewReader(tc.want)
                       } else {
			    p = NewParser(strings.NewReader(tc.want))
                       }
			got, err := p.parseHeader()

			if gotErr := err != nil; gotErr != tc.wantErr {
				t.Fatalf("incorrect error returned from parseHeader, wantErr: %v, err: %v", tc.wantErr, err)
			}

			if diff := cmp.Diff(tc.want, got); !tc.wantErr && diff != "" {
				t.Fatalf("parseHeader returned incorrect header, (-want, +got): %v", diff)
			}

		})
	}
```

This should do 2 very nice things:
1. Contributors don't need to to spend time aligning multi line string literals in accordance to the spec. All they have to do is define a Struct. More complex structs of the Genbank could have some constructors with some nice defaults to help build up the test data if it is too complex to define the struct in place or if we find that it takes too much boiler plate to define over and over again. This should make it easier to add and reason about test data.
2. This gets us part of the way to implementing writing to disk! 

This is probably overkill for header and locus, but maybe good to keep in mind for the future. Having golden tests would be super helpful once we're ready to test some complete Genbank files. This would make it trivial to add test cases and work through any problems if bugs reported.

io/genbank/parser.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+}
+
+/* KEYWORD PARSING FUNCTIONS */
+
+const locusDateLayout = "02-Jan-2006"
+
+// See Genbank spec 3.4.4
+func (p *Parser) parseLocus(ctx context) error {
```
TwFlem, line 316: Might be good to rename context to ParseContext or something. That way there aren't any collisions with the context package.

io/genbank/parser.go
```diff
@@ -0,0 +1,356 @@
...
+		case eof:
+			return res, true, nil
+		case subKeyword:
+			return res, false, p.makeSyntaxError("subkeyword found outside of keyword context")
+		case entryEnd:
+			return res, false, nil
+		case continuation:
+			return res, false, p.makeSyntaxError("continuation found outside of keyword context")
+		case featureKey:
+			return res, false, p.makeSyntaxError("feature key found outside of FEATURES keyword context")
```
TwFlem, line 304: These static errors would be common across the different contexts right? Maybe it is a bit premature, but it might be nice to have a default case with something like

```go
func GetWrongContextError(lineType LineType) error {
    switch lineType {
    // the things
    default: 
        // something went really wrong 
    }
}

That way it can be shared for the other parse functions for the other contexts.

File: io/genbank/parser.go [hunk 4]
```diff
@@ -0,0 +1,356 @@
...
+		return res, fmt.Errorf("failed to read file statistics: %w", err)
+	}
+	if len(releaseLine) < 48 {
+		return res, p.makeSyntaxError("release line too short")
+	}
+	res.NumEntries, err = strconv.Atoi(strings.TrimSpace(statsLine[:8]))
+	if err != nil {
+		return res, p.makeSyntaxError("failed to parse number of entries/loci", err)
+	}
+	res.NumBases, err = strconv.Atoi(strings.TrimSpace(statsLine[15:26]))
```
TwFlem, line 255: The spec has a lot of opinions about spacing. I'm not sure what the best way to handle this is. Perhaps this kind of thing is the best approach. Maybe it could help to abstract this a bit? This could help with reading why the numbers are there, but more importantly perhaps it can also be used for writing? 

For instance, maybe something like 

type TokenRange struct {
  start int
  end int
}

var fileNameRange = TokenRange{15,26}
var statsRange =  TokenRange{39,47}

that way slicing could look like

statsLine[statsRange.start:statsRange.end]

Maybe it is helpful for formatting text for writing as well? 

Just spit balling.

io/genbank/parser.go
carreter: Good catch. Done.

io/genbank/parser.go
carreter: Great idea! Done.

carreter: Regardless, I think it's better for individual functions to bump up to the main `parseEntry` loop if they encounter a linetype they don't recognize. That way, error handling is all in one place.

io/genbank/parser.go
carreter: Absolutely! Good point re: being helpful for writing. Will go back and do this.

io/genbank/parser_test.go
carreter: Good points. We need write ability - makes sense to implement this along the way so we can just do round-trip tests! Will go back and do this.

io/genbank/parser.go
TwFlem: Sounds good to me!

io/genbank/parser.go
carreter: Done now. I think I got all (or most of) the magic numbers. Proved useful with the header write ability!

io/genbank/parser_test.go
carreter: Latest commit implements these suggestions for the header:

- separate "happy" and "sad" tests into `TestXXXRoundtrip` and `TestXXXErr`
- make tests roundtrip instead

Still need to do this for the rest of the spec though!

TwFlem: Lgtm! I'd say ship it!

PR #440, opened 2024-01-19
Run length bwt

TwFlem:
## Changes in this PR

This PR is a memory optimization for the BWT. This implements the RLBWT, Run-Length Burrows Wheeler Transform. The difference between the RLBWT and the BWT is that instead of having the whole last column (the actual BWT string), we have a run length compressed version of the last column. For example, the Last column of banana is: "annb$aa". It's run length compression is "anb$a".

The RLBWT is very effective for repetitive texts like genomic data. For example, in the test data we have for the BWT, we repeat the base string 3 times. The resulting Last column has a length of 340 in the normal BWT implementation. In this implementation, it only has a length of 94 (along with some negligible auxiliary info about run start positions and character cumulative counts). 

Ben would do a much better job at explaining the mechanics and O complexity than I:
* [Nature of Repetitive texts and the BWT](https://www.youtube.com/watch?v=Zf6r8XdYf-M&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&index=18)
* [Algorithm of the RLBWT](https://www.youtube.com/watch?v=xT21EoBmmAc&list=PL2mpR0RYFQsADmYpW2YWBrXJZ_6EL_3nu&index=19)

This is a needed change for the BWT to be a viable tool for real workloads. There will be two follow ups after this PR:

1. reducing the size of the suffix array using an r-index (this will also decrease consumed memory by an order of magnitude)
2. serializing the BWT to disk as a cache. Calculating the BWT takes a lot of up front effort, but can save a lot of time in the long run if the workload involves searching through the BWT. We can this tool even more valuable for Poly's users by allowing them to opt into a way to serialize the BWT so they can write it to disk.

No

## Pre-merge checklist

PS: I didn't update the changelog since the BWT is still unreleased at the time of writing.


search/bwt/bitvector.go
```diff
@@ -72,6 +72,18 @@ func (b bitvector) len() int {
 	return b.numberOfBits
 }
 
+func (b bitvector) toBinStr() string {
```
TwFlem, line 75: This was helpful for debugging. Happy to keep it in, also happy to remove it.

search/bwt/bwt.go
```diff
@@ -269,7 +291,33 @@ func (bwt BWT) Len() int {
...
+		}
+		for lastColumn.Len() < currCharEnd {
+			lastColumn.WriteByte(currChar)
+		}
+	}
+	return lastColumn.String()
+}
+
+func (bwt BWT) getFirstColumnStr() string {
+	firstColumn := strings.Builder{}
```
TwFlem, line 312: This is helpful for debugging. I recommend keeping it for the future.

search/bwt/rsa_bitvector.go
```diff
@@ -97,6 +97,10 @@ func (rsa rsaBitVector) Access(i int) bool {
 	return rsa.bv.getBit(i)
 }
 
+func (rsa rsaBitVector) getBinStr() string {
+	return rsa.bv.toBinStr()
```
TwFlem, line 101: This was helpful for debugging. Happy to keep it in, also happy to remove it.

search/bwt/wavelet.go
```diff
@@ -178,7 +178,10 @@ func (wt waveletTree) Rank(char byte, i int) int {
 	}
 
 	curr := wt.root
-	ci := wt.lookupCharInfo(char)
+	ci, ok := wt.lookupCharInfo(char)
```
TwFlem, line 181: In the OG BWT, we actually never used WaveletTree's select. It is semantically correct to return 0 on !ok. This can happen with a search pattern containing a character that does not exist in the original sequence.

TimothyStiles (changes requested):
I like the debugging functions and the lint directive but without some example usage I'm not sure how they'd be used to debug and would probably end up removing them in the future if they were merged without them.

Think you could write some tiny tests just to demo their usage? You can just put each example in a simple non-example test and refer to it in-line with something like, "To see usage example check out `TestDebugRelatedFunction` in `bwt_test.go`.

search/bwt/bitvector.go
TwFlem: Removed :)

search/bwt/rsa_bitvector.go
TwFlem: Removed :)

search/bwt/bwt.go
TwFlem: @TimothyStiles I added an explanation and demo in this commit: https://github.com/bebop/poly/pull/440/commits/fe3b686a7907423a540ca11ef50bcc5f74de0810#diff-c08e2ebed9f275352a42178cdcc0b15abc0ca64e019428e5fe8b6ff5ea3af188R636

Also, I'm not sure if we have any better way of switching on debug for Poly's internals (didn't find anything), but I added that debug field in the BWT. LMK if there's another approach that is preferred or if the explanation along with the lint ignore annotation is good enough.

PR #447, opened 2024-09-20
Separated minimums between design and simulate

whitead:
## Changes in this PR

The minimum primer length is set at 15, which is probably reasonable for designing new primers but is too restrictive for simulating given primers. The cited reference in code comment supports using 7, instead of 15. 

This changes there to be a minimum for design (15, as existing code uses) and for simulate (7, as cited in paper). I also added an error for passing in short primers.

I would like to use simulate to test my own primers, but index errors are thrown if I use primers smaller than minimalPrimerLength

The design code is unchanged. The simulate behavior changes because if someone simulated with short primers, the code would previously blow up with index error.

## Pre-merge checklist

I actually do not see what linter is used (I've not written Go code before) - so please let me know what is appropriate here! Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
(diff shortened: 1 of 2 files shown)


TimothyStiles (commented):
This looks great @whitead! Can't believe I missed this in my notifications!

The CI/CD workflows for linting and testing don't seem to be triggering but that's on me. I'll fix that up and get this merged.

